PR #4, opened 2021-08-19
Fix tests

irinamihai:
Description:
- Fix assert YAMLs for kuttl tests to include the new names for the
  ACM objects
- Update Makefile to build new docker image from current source tree
  and load it to the test kind cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.0.2
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "preview,fast,stable")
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -126,6 +126,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	kind load docker-image ${IMG} --name ${KIND_NAME}
```
rcarrillocruz, line 129: You should not add this on the deploy target.
make deploy is the usual command used to deploy an operator to a working cluster:

https://sdk.operatorframework.io/docs/building-operators/golang/tutorial/#2-run-as-a-deployment-inside-the-cluster

If you put this here, it assumes it's on a kind environment.

It's better if you add a new 'kind-load-operator-image' that does the kind load command and then you call this new target on the complete-kuttl-test

rcarrillocruz:
Since you are fixing the Makefile, how about you also change IMAGE_TAG_BASE to point to quay.io/openshift-kni/cluster-group-upgrades-operator ?

Makefile
irinamihai: Yes, you are right. I'll make the change. Thanks!

Makefile
```diff
@@ -234,6 +234,10 @@ kind-create-cluster:
 kind-delete-cluster:
 	kind delete cluster --name $(KIND_NAME)
 
+kind-prepare-cluster-for-tests:
```
rcarrillocruz, line 237: I find this name a bit odd, why not just kind-load-operator-image , so it's more explicit?

rcarrillocruz:
Btw, you can add me to 'reviewer' to PRs, that way I don't have to dig into repos to see the list. It's ok now, but it will become a bit unmanageable if we start having more PRs to review.

OH , and finally I made you owner of the repo, hope you got the invite in your inbox.

Makefile
irinamihai: I was thinking in the future we might need to do other actions to prepare the cluster for running the tests.

rcarrillocruz: i think we better be more focused on names, just like the other targets. If we plan to do more things, then we can generalize/add new targets.

irinamihai: Ok, sure.

rcarrillocruz (approved):

PR #14, opened 2021-09-13
Add deleteObjectsOnCompletion to spec

rcarrillocruz:
This will delete ACM objects on upgrade completion if set

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -151,7 +151,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 						if err != nil {
 							return ctrl.Result{}, err
 						}
-						r.Log.Info("Set remediationAction to enforce on Policy object", "policy", policy.GetName())
+						r.Log.Info("Set remediationAction to enforce on Policy", "policy", policy.GetName())
 					}
 
 					statusObject := policy.Object["status"].(map[string]interface{})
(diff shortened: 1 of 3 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -224,20 +224,24 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		}
-		r.deletePolicies(ctx, clusterGroupUpgrade)
-		if err != nil {
-			return ctrl.Result{}, err
+		if clusterGroupUpgrade.Spec.DeleteObjectsOnCompletion {
+			err := r.deletePlacementRules(ctx, clusterGroupUpgrade)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			r.deletePlacementBindings(ctx, clusterGroupUpgrade)
```
irinamihai, line 237: err :=
?

controllers/clustergroupupgrade_controller.go
```diff
@@ -224,20 +224,24 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					Message: "The ClusterGroupUpgrade CR has all upgrade policies compliant",
 				})
 			}
+		} else if readyCondition.Reason == "UpgradeTimedOut" {
+
```
irinamihai, line 228: Was wondering if we could add a Log here, just to get an idea of what's happening. Your choice if it makes sense.

controllers/clustergroupupgrade_controller.go
rcarrillocruz: yep, you are right, thx!

yrobla (approved):

PR #24, opened 2021-09-30
Update install namespace and create manged clusters for KUTTL tests

irinamihai:


controllers/clustergroupupgrade_controller.go
```diff
@@ -1043,11 +1043,12 @@ func (r *ClusterGroupUpgradeReconciler) validateCR(ctx context.Context, clusterG
...
 		})
 		err := r.Client.Get(ctx, client.ObjectKey{
-			Name: cluster,
+			Name:      cluster,
+			Namespace: cluster,
 		}, foundManagedCluster)
 
 		if err != nil {
-			return fmt.Errorf("Cluster %s is not a ManagedCluster", cluster)
+			return fmt.Errorf("Cluster %s is not a ManagedCluster!!!", cluster)
```
irinamihai, line 1051: Remove this.

rcarrillocruz: It seems you are adding exclamations, yet you self-commented to remove them? Is it 'reminder' you put yourself to address it shortly?

deploy/upgrades/canaries/clustergroupupgrade.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: ran.openshift.io/v1alpha1
 kind: ClusterGroupUpgrade
 metadata:
   name: canaries
-  namespace: default
+  namespace: openshift-cluster-group-upgrades
```
rcarrillocruz, line 5: Why you change all the namespaces of the CRs for testing?
The 'openshift-cluster-group-upgrades' namespace is supposedly where the controller will be placed unto, but the user can and should be able to open ClusterGroupUpgrade CRs on any namespace.

controllers/templates.go
```diff
@@ -68,7 +68,7 @@ spec:
             exclude:
               - kube-*
             include:
-              - default
+              - openshift-cluster-group-upgrades
```
rcarrillocruz, line 71: you don't need to add this. THe namespace openshift-cluster-group-upgrades is a namespace where the controller runs, which is on the hub. This namespace selector field is on configuration policies, which are meant on the spokes.
As a matter of fact, this 'default' is a bug, I should probably change to * .

PR #35, opened 2021-11-09
Update ClusterGroupUpgrades CRD

irinamihai:
Description:
- Update ClusterGroupUpgrades CRD to include ManagedPolicy and Enable in the ClusterGroupUpgradeSpec
- Update existing tests with dummy entries for the new members of the CRD
- Note: Current CRD members haven't been removed. We'll cleanup the CRD members that we won't be using (like PlatformUpgrade and OperatorUpgrade) as we're removing the functionality from the operator.
- KUTTL tests have been run locally.

### DIFF
--- a/api/v1alpha1/clustergroupupgrade_types.go
+++ b/api/v1alpha1/clustergroupupgrade_types.go
@@ -61,6 +61,9 @@ type ClusterGroupUpgradeSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	Clusters            []string                 `json:"clusters,omitempty"`
 	RemediationStrategy *RemediationStrategySpec `json:"remediationStrategy,omitempty"`
+	ManagedPolicy       []string                 `json:"managedPolicy,omitempty"`
+	//+kubebuilder:default=false
+	Enable bool `json:"enable,omitempty"`
 	//+kubebuilder:validation:Enum=inform;enforce
 	//+kubebuilder:default=inform
 	RemediationAction         string                `json:"remediationAction,omitempty"`
--- a/api/v1alpha1/clustergroupupgrade_types.go
+++ b/api/v1alpha1/clustergroupupgrade_types.go
@@ -61,7 +61,7 @@ type ClusterGroupUpgradeSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	Clusters            []string                 `json:"clusters,omitempty"`
 	RemediationStrategy *RemediationStrategySpec `json:"remediationStrategy,omitempty"`
-	ManagedPolicy       []string                 `json:"managedPolicy,omitempty"`
+	ManagedPolicies     []string                 `json:"managedPolicies,omitempty"`
 	//+kubebuilder:default=false
 	Enable bool `json:"enable,omitempty"`
 	//+kubebuilder:validation:Enum=inform;enforce
--- a/api/v1alpha1/clustergroupupgrade_types.go
+++ b/api/v1alpha1/clustergroupupgrade_types.go
@@ -59,17 +59,17 @@ type OperatorUpgradeSpec struct {
 type ClusterGroupUpgradeSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	Clusters            []string                 `json:"clusters,omitempty"`
-	RemediationStrategy *RemediationStrategySpec `json:"remediationStrategy,omitempty"`
-	ManagedPolicies     []string                 `json:"managedPolicies,omitempty"`
-	//+kubebuilder:default=false
-	Enable bool `json:"enable,omitempty"`
 	//+kubebuilder:validation:Enum=inform;enforce
 	//+kubebuilder:default=inform
-	RemediationAction         string                `json:"remediationAction,omitempty"`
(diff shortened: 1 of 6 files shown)


browsell (commented):
Don't see this in the diff 
- Remove the members that we won't be using (like PlatformUpgrade
 and OperatorUpgrade) as we're removing the functionality from the operator.

irinamihai:
> Don't see this in the diff
> 
> * Remove the members that we won't be using (like PlatformUpgrade
>   and OperatorUpgrade) as we're removing the functionality from the operator.

@browsell , this is just a note explaining why I left the other members as the operator functionality hasn't changed yet.

api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -61,6 +61,9 @@ type ClusterGroupUpgradeSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	Clusters            []string                 `json:"clusters,omitempty"`
 	RemediationStrategy *RemediationStrategySpec `json:"remediationStrategy,omitempty"`
+	ManagedPolicy       []string                 `json:"managedPolicy,omitempty"`
+	//+kubebuilder:default=false
+	Enable bool `json:"enable,omitempty"`
```
serngawy, line 66: can you apply `$go fmt ` it will fix the style

irinamihai: Heh, I did and that's what it did. I think it's like that because it's between //+kubebuilder markers.

imiller0, line 66: I will defer to standard practices for CRDs, but it seems like "enable" is a field that always has a value and should always be represented in the output. Should we remove the omitempty from this field?

api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -61,6 +61,9 @@ type ClusterGroupUpgradeSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	Clusters            []string                 `json:"clusters,omitempty"`
 	RemediationStrategy *RemediationStrategySpec `json:"remediationStrategy,omitempty"`
+	ManagedPolicy       []string                 `json:"managedPolicy,omitempty"`
```
imiller0, line 64: Since this field is a list should the field name be plural? ie 
```suggestion
	ManagedPolicies     []string                 `json:"managedPolicies,omitempty"`
```

imiller0 (changes requested):
Should we rename this CRD to better reflect that it can be used beyond simple upgrades (eg it can manage changes to CRs or any other "staged rollout" of policies).

Are the three ClusterGroupUpgrade CRs reference implementations or test data or something else? Their location (under deploy/...) doesn't make clear (to me) their purpose.

irinamihai:
> Should we rename this CRD to better reflect that it can be used beyond simple upgrades (eg it can manage changes to CRs or any other "staged rollout" of policies).
> 
> Are the three ClusterGroupUpgrade CRs reference implementations or test data or something else? Their location (under deploy/...) doesn't make clear (to me) their purpose.

I did think about renaming the CRD, but my concern is we'll have to rename it again once we're agreed on a final name for the operator. I'd say to keep it as is for the time being.

The deploy directory is used for any CRDs that we need for deploying a cluster with or without tests. I actually followed the folder structure from governance-policy-propagator: https://github.com/open-cluster-management/governance-policy-propagator/tree/main/deploy. I'm happy to rename and move stuff around for a more intuitive structure.

@imiller0 , kindly let me know what you think about these.

api/v1alpha1/clustergroupupgrade_types.go
irinamihai: Right now all the members are using "omitempty" even if they are always used, apparently, but I agree with what you're saying. My proposal is to have a cleanup ticket to address these types of issue + any cleanup in the controller code once the restructuring functionality is in place.

serngawy: I believe if you re-order the struct fields having fields with annotation in top fmt will align them all.

imiller0:
> I did think about renaming the CRD, but my concern is we'll have to rename it again once we're agreed on a final name for the operator. I'd say to keep it as is for the time being.
I do think we should rename prior to feature-freeze but I'm OK with renaming in a follow-on PR.

> The deploy directory is used for any CRDs that we need for deploying a cluster with or without tests. I actually followed the folder structure from governance-policy-propagator: https://github.com/open-cluster-management/governance-policy-propagator/tree/main/deploy. I'm happy to rename and move stuff around for a more intuitive structure.
Ok, so these files are used for deploying a cluster for testing?

irinamihai:
> > I did think about renaming the CRD, but my concern is we'll have to rename it again once we're agreed on a final name for the operator. I'd say to keep it as is for the time being.
> 
> I do think we should rename prior to feature-freeze but I'm OK with renaming in a follow-on PR.
Will add to cleanup ticket.

> > The deploy directory is used for any CRDs that we need for deploying a cluster with or without tests. I actually followed the folder structure from governance-policy-propagator: https://github.com/open-cluster-management/governance-policy-propagator/tree/main/deploy. I'm happy to rename and move stuff around for a more intuitive structure.
> 
> Ok, so these files are used for deploying a cluster for testing?
Yes. Either for testing or development.

nishant-parekh (approved):

PR #37, opened 2021-11-23
Update cluster selection criteria

irinamihai:
Description:
- add function to find out which clusters from a ClusterGroupUpgrade CR are NonCompliant with all the managedPolicies from that same CR: function returns a map of type map[string]bool which will hold true if a cluster is NonCompliant will all the managedPolicies in the CR and false if a cluster is NOT NonCompliant with all the managedPolicies in the CR
- only the clusters obtained above which are NonCompliant with all the mangedPolicies are considered for the remediation plan

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -100,7 +100,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			// Before starting the upgrade check that all the managed policies exist.
 			allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent := r.doManagedPoliciesExist(ctx, clusterGroupUpgrade)
 			if allManagedPoliciesExist == true {
-				r.buildRemediationPlan(ctx, clusterGroupUpgrade)
+				r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
 				err := r.reconcileResources(ctx, clusterGroupUpgrade, managedPoliciesPresent)
 				if err != nil {
 					return ctrl.Result{}, err
@@ -324,6 +324,9 @@ func (r *ClusterGroupUpgradeReconciler) copyManagedInformPolicyForBatch(
 
 	// Go through the managedPolicies in the CR and make sure they exist.
 	for managedPolicyIndex, managedPolicy := range policies {
+		// Check that at least one cluster in the batch is NonCompliant with the managedPolicy.
+		//clustersNonCompliantWithPolicy := getClustersNonCompliantWithPolicy(managedPolicy)
+
 		// Create a new unstructured variable to keep all the information for the new policy.
 		newPolicy := &unstructured.Unstructured{}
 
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
+		} else {
+			clusterMap[clusterName] = false
+		}
+	}
+	r.Log.Info("Cluster (non) compliant with all managedPolicies map  : ", "clustersNotCompliantMap", clusterMap)
+	return clusterMap
+}
+
+func (r *ClusterGroupUpgradeReconciler) buildRemediationPlan(ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, managedPolicies []*unstructured.Unstructured) {
+	clusterNonCompliantWithManagedPoliciesMap := r.getClustersNonCompliantWithAllManagedPolicies(ctx, clusterGroupUpgrade, managedPolicies)
```
imiller0, line 1041: Does this mean that we only remediate clusters where ALL the managedPolicies are non compliant? My understanding is that the operator needs to remediate any cluster where any managedPolicy is non compliant.

controllers/clustergroupupgrade_controller.go
```diff
@@ -990,16 +1069,22 @@ func (r *ClusterGroupUpgradeReconciler) buildRemediationPlan(ctx context.Context
...
-			if !isCanary[site] {
-				batch = append(batch, site)
-			}
+
+	var batch []string
+	batchCount := 0
+	for i := 0; i < len(clusters); i++ {
+		site := clusters[i]
+		if !isCanary[site] && clusterNonCompliantWithManagedPoliciesMap[site] == true {
+			batch = append(batch, site)
```
imiller0, line 1078: When there is a list of policies to be remediated does the operator remediate the policies in the order they are listed?

controllers/clustergroupupgrade_controller.go
browsell: The operator needs to reconcile any non-compliant policy.

controllers/clustergroupupgrade_controller.go
browsell: It will need to remediate in the order they are listed in the CR and wait until the previous one is remediate before continuing.

controllers/clustergroupupgrade_controller.go
irinamihai: Thank you for pointing this out, @imiller0. I'll update this.

controllers/clustergroupupgrade_controller.go
irinamihai: Yes. I have a separate ticket to update the reconcile logic for doing that.

controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
+			// increment the number of policies for which it's NonCompliant.
+			if crtSubStatus["compliant"] == "NonCompliant" {
+				if _, ok := clustersNonCompliant[crtSubStatus["clustername"].(string)]; ok {
+					clustersNonCompliant[crtSubStatus["clustername"].(string)]++
+				}
+			}
+		}
+	}
+
+	for clusterName, clusterNonCompliantNumber := range clustersNonCompliant {
```
serngawy, line 1029: I believe its better to return a array/slice of the clustersNonCompliant instead of all clusters in map[string]bool so when you call this func u shorter list and don't need to check for the boolean.

controllers/clustergroupupgrade_controller.go
serngawy: I believe so as well, the operator need to reconcile any non-compliant policy with the given clusters. Is there a reason that will make issue ?

controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
-			remediationPlan = append(remediationPlan, []string{canary})
-			isCanary[canary] = true
+			// TODO: make sure the canary clusters are in the list of clusters.
+			if clusterNonCompliantWithManagedPoliciesMap[canary] == true {
+				remediationPlan = append(remediationPlan, []string{canary})
+				isCanary[canary] = true
+			}
 		}
 
 	}
```
serngawy, line 1065: Kind of strange logic or I miss understand; what make a cluster exist in both `clusterGroupUpgrade.Spec.RemediationStrategy.Canaries` AND `clusterGroupUpgrade.Spec.Clusters` why there are 2 for loops to identify if the cluster is canary ? if it is a canary cluster so it will be in the canary list if not then it will be in the cluster list.

controllers/clustergroupupgrade_controller.go
serngawy: I do remember we said that we will assign a priority/order value to the policy ? default is 0 for all if there is priority/order so it needs to be defined. May be future enhancement ?

serngawy (changes requested):

controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
+
+	for _, managedPolicy := range managedPolicies {
+		statusObject := managedPolicy.Object["status"].(map[string]interface{})
+		statusCompliance := statusObject["status"]
+
+		if statusCompliance == nil {
+			continue
+		}
+
+		for key, value := range statusObject {
```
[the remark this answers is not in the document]
irinamihai: It was for debug, I'll remove.

controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
+			r.Log.Info("key and value", "key", key, "value", value)
+		}
+
+		subStatus := statusCompliance.([]interface{})
+
+		for subStatusIndex := range subStatus {
+			crtSubStatus := subStatus[subStatusIndex].(map[string]interface{})
```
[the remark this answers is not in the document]
irinamihai: I see, yes, you're right. Thanks! I'll update to:
`for _, crtSubStatusCrt := range subStatus {
	crtSubStatusMap := crtSubStatusCrt.(map[string]interface{})`
@serngawy , do you agree?

controllers/clustergroupupgrade_controller.go
```diff
@@ -973,14 +973,93 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
+			// If the cluster is NonCompliant in this current policy and it's also present in the list of clusters from the CR
+			// increment the number of policies for which it's NonCompliant.
+			if crtSubStatus["compliant"] == "NonCompliant" {
```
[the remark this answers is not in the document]
irinamihai: Yes, agreed.

controllers/clustergroupupgrade_controller.go
irinamihai: Yes, it's actually how I've done it initially but thought it would be nice to have a view of what clusters from the spec are also compliant. I can change it back.

controllers/clustergroupupgrade_controller.go
irinamihai: This is actually the initial implementation which contained all the clusters that we want to update in the `clusterGroupUpgrade.Spec.Clusters` and if from that list we want any of them to be canaries, we'll put them in `clusterGroupUpgrade.Spec.RemediationStrategy.Canaries`. I haven't changed that initial logic.

irinamihai:
>I do remember we said that we will assign a priority/order value to the policy ? default is 0 for all if there is priority/order so it needs to be defined. May be future enhancement ?
@serngawy The final decision was that we'll apply the policies in the order they're present in the managedPolicies field. No priorities for now.

nishant-parekh:
@irinamihai can we close this PR? I think [this](https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/39) change covers the changes made here.

nishant-parekh (commented):
@irinamihai can we close this PR? I think this change covers the changes made here.

PR #40, opened 2021-12-05
Add image pre-cache feature

vitus133:
This commit adds a container image and reference manifests for OCP release OLM image spec formation and pre-caching.
The solution is not yet integrated with the cluster-group-upgrade-operator
/cc @browsell @irinamihai @lack @nishant-parekh


pre-cache/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM registry.access.redhat.com/ubi8-minimal:latest
```
browsell, line 1: Since this will be bundled with the operator, we should just include under the existing Containerfile

pre-cache/Makefile
```diff
@@ -0,0 +1,30 @@
+IMG ?= quay.io/vgrinber/pre-cache
```
browsell, line 1: Include under existing Makefile

pre-cache/spoke-deploy/cm.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
browsell, line 1: Include under existing config dir

pre-cache/spoke-deploy/job.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: batch/v1
```
browsell, line 1: See above comment

Dockerfile
```diff
@@ -19,9 +19,19 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:nonroot as operator_image
```
nishant-parekh, line 22: I am not sure how we can use reference this downstream.. is there an alternative to this image that would work for you?

vitus133: I didn't change this part - it was there from the beginning. I only added "as operator_image". We probably will need to change it to one of the UBIs if required by the downstream process, but this is not related to this PR

Dockerfile
```diff
@@ -19,9 +19,19 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
...
 WORKDIR /
 COPY --from=builder /workspace/manager .
```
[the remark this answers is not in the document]
vitus133: FROM golang:1.15 as builder
At line 2

Dockerfile
nishant-parekh: I see my bad.. I will try and figure how this is handled downstream.

Dockerfile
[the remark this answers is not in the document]
nishant-parekh: ohh sorry got it.. so the question is that this Dockerfile is building the operator image.. as for the precache_image as we are independently releasing it as an independent image do we need a separate Dockerfile for it?

Makefile
```diff
@@ -37,6 +37,8 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
+PRECACHE_IMG ?= $(IMAGE_TAG_BASE)-precache:$(VERSION)
```
imiller0, line 40: Does VERSION have a default value if not specified by the caller?

config/pre-cache/cm.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ConfigMap
+metadata:
+  annotations:
+    kubernetes.io/description: Contains a pre-upgrade artifact caching specification
+  name: pre-cache-spec
+  namespace: pre-cache
+data:
+  platform.image: quay.io/openshift-release-dev/ocp-release:4.9.4-x86_64
+  operators.indexes: |
+    dhcp16-231-198.ptp.lab.eng.bos.redhat.com:5000/custom-redhat-operators:1.0.0
```
imiller0, line 11: Please use generic example servers. Maybe "registry.example.com:5000"

config/pre-cache/job.yaml
```diff
@@ -0,0 +1,80 @@
...
+spec:
+  backoffLimit: 0
+  activeDeadlineSeconds: 14400
+  template:
+    metadata:
+      name: pre-cache
+    spec:
+      containers:
+      - name: pre-cache-container
+        image: quay.io/vgrinber/pre-cache:0.0.3
```
imiller0, line 14: Can this be made generic?

config/pre-cache/job.yaml
```diff
@@ -0,0 +1,80 @@
...
+        env:
+        - name: pull_timeout
+          value: "14400"
+        - name: config_volume_path
+          value: /etc/config
+        command:
+        - /bin/bash
```
imiller0, line 21: Can you make a single entry point executable (can be a shell script) which includes all the commands specified here. That allows future versions of the container to update this logic without requiring an update to the Jobs.

config/pre-cache/job.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: config-volume
+          mountPath: /etc/config 
+        - name: registries
+          mountPath: /etc/containers/registries.conf
+        - name: policy
+          mountPath: /etc/containers/policy.json
+        - name: etcdocker
+          mountPath: /etc/docker
+        - name: usr
+          mountPath: /usr
```
imiller0, line 44: Can any of these be mounted read-only?

imiller0 (commented):
Are there existing tools for mirroring images, working with olm indexes, etc that can be used instead of maintaining our own?

Dockerfile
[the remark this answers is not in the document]
nishant-parekh: so we would only release the operator and the pre cache image would be just an image in the bundle for the operator. Is that understanding correct?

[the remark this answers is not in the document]
browsell: yes

Makefile
vitus133: Yes, it's 0.0.3 (line 6)

config/pre-cache/job.yaml
vitus133: Done

vitus133:
> Are there existing tools for mirroring images, working with olm indexes, etc that can be used instead of maintaining our own?

@imiller0 This would be the opm tool. I am still struggling with it trying to use it for rendering the OLM index. So far I can render sqlite index.db file with it, but not a locally stored index image. Even if I do - no confidence it will work equally well with the fs-based index images. I have a story for that (CNF-3669).

config/pre-cache/job.yaml
vitus133: Done

config/pre-cache/job.yaml
vitus133: Done

config/pre-cache/cm.yaml
vitus133: Done

vitus133:
@imiller0 the latest push (f0b33f9) removed sqlite parsing code, and instead is using opm binary to export OLM index images onto the file system.

config/pre-cache/cm.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
browsell, line 1: I assume this is just an example and that the UO would generate this configmap based on the contents of the manged policies ?

PR #42, opened 2021-12-06
Add clusterSelector

irinamihai:
Description:
- add new clusterSelector label to CRD
- adapt controller logic to get all the clusters matching the requested label(s) and include them when computing the remediation
  plan
- ensure a cluster doesn't end up more than once in the final list of clusters

### DIFF
--- a/config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
+++ b/config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
@@ -37,6 +37,16 @@ spec:
           spec:
             description: ClusterGroupUpgradeSpec defines the desired state of ClusterGroupUpgrade
             properties:
+              clusterSelector:
+                description: 'This field holds a label common to multiple clusters
+                  that will be updated. The expected format is as follows: clusterSelector:   -
+                  label1Name=label1Value   - label2Name=label2Value If the value is
+                  empty, then the expected format is: clusterSelector:   - label1Name
+                  All the clusters matching the labels specified in clusterSelector
+                  will be included in the update plan.'
+                items:
+                  type: string
+                type: array
               clusters:
                 items:
                   type: string
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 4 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -101,9 +102,13 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent := r.doManagedPoliciesExist(ctx, clusterGroupUpgrade)
 			if allManagedPoliciesExist == true {
 				// If all the managedPolicies exist, continue with building the upgrade batches.
-				r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
+				err := r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
```
vitus133, line 105:
```suggestion
				err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
```
err is already defined in the outer scope, can it be reused here?

vitus133 (commented):
Hi, just a minor comment there. Otherwise looks good!

config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
```diff
@@ -37,6 +37,10 @@ spec:
           spec:
             description: ClusterGroupUpgradeSpec defines the desired state of ClusterGroupUpgrade
             properties:
+              clusterSelector:
```
imiller0, line 40: Including a description here of the structure/format of the string would be really helpful.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1012,12 +1030,64 @@ func (r *ClusterGroupUpgradeReconciler) buildRemediationPlan(ctx context.Context
...
 
 func getResourceName(clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, policyName string) string {
 	return clusterGroupUpgrade.Name + "-" + policyName
 }
 
+func (r *ClusterGroupUpgradeReconciler) getAllClustersForUpgrade(ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) ([]string, error) {
+	var clusterNames []string
+	keys := make(map[string]bool)
+	for _, clusterSelector := range clusterGroupUpgrade.Spec.ClusterSelector {
+		selectorList := strings.Split(clusterSelector, "=")
```
imiller0, line 1045: Including a comment here with the format of the clusterSelector string would be helpful in understanding the logic.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1012,12 +1030,64 @@ func (r *ClusterGroupUpgradeReconciler) buildRemediationPlan(ctx context.Context
...
+		r.Log.Info("selectorList", "selectorList", selectorList)
+		var clusterLabels = make(map[string]string)
+		if len(selectorList) == 2 {
+			clusterLabels = map[string]string{strings.Split(clusterSelector, "=")[0]: strings.Split(clusterSelector, "=")[1]}
```
imiller0, line 1049: minor nit: do the split once and then choose how you index into/use the result.

imiller0 (changes requested):
A couple of minor nits but overall looks good. 
Please add unit tests for the new functionality before merging.

controllers/clustergroupupgrade_controller.go
irinamihai: Yes, thank you :+1:

PR #43, opened 2021-12-08
Use inform policies

irinamihai:
Description:

- use the inform policies to decide both the clusters from the remediation plan and the next policy to enforce for a cluster
- when we are upgrading a cluster we start with the first policy for which the cluster is NonCompliant. Once the cluster is
  Compliant with a policy, we move to the next one for which the cluster is NonCompliant and so on until the batch is completed
- to decide of a cluster is Compliant or NonCompliant with an informed policy, we look for the cluster name in the policy's
  status.status. All the clusters present in the policy's placement rule should appear there, either as Compliant or as NonCompliant
- if an Upgrade timed out, check if the batch finished meanwhile. We are doing this by checking if the clusters in the batch
  (which should be the last batch for this check to even make sense) are Compliant or not matched with the remaining managed policies
- add short name for the CR: cgu

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -94,28 +94,28 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			Type:    "Ready",
 			Status:  metav1.ConditionFalse,
 			Reason:  "UpgradeNotStarted",
-			Message: "The ClusterGroupUpgrade CR has remediationAction set to inform",
+			Message: "The ClusterGroupUpgrade CR is not enabled",
 		})
 	} else if readyCondition.Status == metav1.ConditionFalse {
 		if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == "UpgradeCannotStart" {
 			// Before starting the upgrade check that all the managed policies exist.
 			allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent := r.doManagedPoliciesExist(ctx, clusterGroupUpgrade)
 			if allManagedPoliciesExist == true {
-				// If all the managedPolicies exist, continue with building the upgrade batches.
-				err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
-				if err != nil {
-					return ctrl.Result{}, err
-				}
-
 				// Create the needed resources for starting the upgrade.
 				err = r.reconcileResources(ctx, clusterGroupUpgrade, managedPoliciesPresent)
 				if err != nil {
 					return ctrl.Result{}, err
 				}
 
+				// Build the upgrade batches.
+				err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+
 				var statusReason, statusMessage string
 				statusReason = "UpgradeNotStarted"
-				statusMessage = "The ClusterGroupUpgrade CR has remediationAction set to inform"
+				statusMessage = "The ClusterGroupUpgrade CR is not enabled"
 				if clusterGroupUpgrade.Spec.Enable == true {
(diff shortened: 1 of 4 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -866,119 +899,95 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
-		if crtSubStatusMap["compliant"] == utils.StatusNonCompliant && clusterName == crtSubStatusMap["clustername"].(string) {
-			return false
-		}
-	}
-
-	return true
-}
+/*
+  getClusterComplianceWithPolicy returns the compliance of a certain cluster with a certain policy.
+  based on a policy's status structure which is below.
```
imiller0, line 904: This would be a good place to capture the information you got from ACM that the status will always include any bound clusters and be either compliant/non-compliant (reference to their API docs is even better)

controllers/clustergroupupgrade_controller.go
```diff
@@ -866,119 +899,95 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
-	statusCompliance := statusObject["status"]
+func (r *ClusterGroupUpgradeReconciler) getClusterComplianceWithPolicy(
+	clusterName string, policy *unstructured.Unstructured) (*string, error) {
 
-	// If all the clusters are compliant, return nil.
-	if statusCompliance == nil {
-		return nil
+	// Get the compliant status part of the policy.
+	statusObject := policy.Object["status"].(map[string]interface{})
+	if statusObject["compliant"] == nil {
```
imiller0, line 927: Under what conditions does the global state go to pending (nil)? Is it possible that the global compliance could be pending but we have a valid status for the cluster we care about? For example what if the inform policy has a cluster which is not part of this upgrade CR that is disconnected from the hub cluster. Would a "pending" state on that one cluster cause the global compliance to be pending as well even though we could still query the substatus for the cluster we are handling?

controllers/clustergroupupgrade_controller.go
```diff
@@ -866,119 +899,95 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
...
 	for _, crtSubStatusCrt := range subStatus {
 		crtSubStatusMap := crtSubStatusCrt.(map[string]interface{})
-		// If the cluster is NonCompliant, add it to the list of NonCompliant clusters.
-		if crtSubStatusMap["compliant"] == utils.StatusNonCompliant {
-			clustersNonCompliant = append(clustersNonCompliant, crtSubStatusMap["clustername"].(string))
+		// If the cluster is Compliant, return true.
+		if clusterName == crtSubStatusMap["clustername"].(string) {
+			if crtSubStatusMap["compliant"] == utils.StatusCompliant {
+				clusterCompliance = utils.StatusCompliant
+				return &clusterCompliance, nil
```
imiller0, line 947: Why return a pointer here rather than a simple constant (compliant vs non-compliant)?

controllers/clustergroupupgrade_controller.go
```diff
@@ -389,38 +383,39 @@ func (r *ClusterGroupUpgradeReconciler) updatePlacementRuleWithCluster(
 			return err
```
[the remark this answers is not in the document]
irinamihai: @imiller0 What other errors do you think would be needed? I included 2: when the placementRule doesn't exist and when it can't be updated. I'm handling the case when the cluster to add is already present in the placementRule.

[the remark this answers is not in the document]
imiller0: I'm not sure what errors the above r.Client.Get(...) call can return but the code `if errors.IsNotFound(err)` qualifier on this `return err` seems to indicate that we would not return an error for other error conditions (maybe an API server timeout or other transient conditions?). Seems odd to get `err != nil` on line 380 and then only return an error for one possible condition. I know that isn't code you touched with this PR but I wanted to flag it for review (either now or a subsequent PR). Hope that helps. Thanks!

PR #44, opened 2021-12-14
Find the namespaces for each managed policy

irinamihai:
Description:
- in the Upgrade CR status keep all the managed policies with their namespaces which are probably different from the default namespace and from the namespace of the Upgrade CR
- in order to find the namespace for each managed policy, look for child policies in the namespaces that have the same name as the clusters from the Upgrade CR. Use the child policy name format to determine the namespace for each managed policy

### DIFF
--- a/config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
+++ b/config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
@@ -155,6 +155,14 @@ spec:
                   - type
                   type: object
                 type: array
+              copiedPolicies:
+                items:
+                  type: string
+                type: array
+              managedPoliciesNs:
+                additionalProperties:
+                  type: string
+                type: object
               placementBindings:
                 description: 'INSERT ADDITIONAL STATUS FIELD - define observed state
                   of cluster Important: Run "make" to regenerate code after modifying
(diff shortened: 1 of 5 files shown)


config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
```diff
@@ -155,6 +155,10 @@ spec:
                   - type
                   type: object
                 type: array
+              managedPoliciesNs:
+                additionalProperties:
+                  type: string
+                type: object
               placementBindings:
```
imiller0, line 162: This status is "disconnected" from the list of managed policies below. Would it make sense to keep the namespaces associated with the policies from which they were derived? For example, making the "policies" status an object:
```
status:
  policies:
     - name: policy1
       ns: policy1NS
     - name: policy2
       ns: policy2NS
```

irinamihai: This list is actually a list of the copied policies. I could make that policies status an object and include the name of the copied policy and the namespace of the parent policy. To be honest, I think the current approach is a bit cleaner, having the managed policies together with their namespaces.

imiller0: Maybe we can do a walk-through of the status and make any improvements we identify in a followup PR?

imiller0 (commented):
/lgtm

PR #46, opened 2021-12-14
Create clusterGroupUpgrade for each managedcluster

Missxiaoguo:
Create a new controller to only watch managedCluster.
When a new managedcluster is created and ready, a ClusterGroupUpgrade
CR for this cluster will be created under namespace "ztp-install" with
the content of cluster name, list of policies and maxConcurrency.

The list of policies is generated from cluster's child policies and
the order is built based on the policy ztp-deploy-wave annotation.


controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ManagedClusterForCguReconciler) getPolicies(ctx context.Context, clusterName string) (*policiesv1.PolicyList, error) {
+	policies := &policiesv1.PolicyList{}
+	if err := r.List(ctx, policies, client.InNamespace(clusterName)); err != nil {
+		return nil, err
+	}
+	return policies, nil
```
irinamihai, line 125: I'm wondering if we need an extra check to make sure these policies are indeed child policies and not another policy that somehow ended up in the same namespace? Child policies have a specific label.

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Create a clusterGroupUpgrade
+func (r *ManagedClusterForCguReconciler) newClusterGroupUpgrade(ctx context.Context, cluster *clusterv1.ManagedCluster, policies *policiesv1.PolicyList) (err error) {
+	var managedPolicies []string
+	var policyWaveMap = make(map[string]string)
+	for _, policy := range policies.Items {
+		deployWave := policy.GetAnnotations()[ztpDeployWaveAnnotation]
+		// apply the policy at the last if policy doesn't have wave
+		if deployWave == "" {
+			deployWave = "1000"
```
irinamihai, line 136: Could we add a comment or link and explain what there wave values means?

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+	Scheme *runtime.Scheme
+}
+
+// Reconcile the managed cluster auto create ClusterGroupUpgrade
+// - When a new managed cluster is created, create ClusterGroupUpgrade CR for the cluster only when it's ready
+//   and its child policies are available
+//
+// Note: The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ManagedClusterForCguReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
imiller0, line 64: It would be helpful to document the conditions which cause this method to get called.

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	availableCondition := meta.FindStatusCondition(managedCluster.Status.Conditions, clusterv1.ManagedClusterConditionAvailable)
+	if availableCondition == nil {
+		r.Log.Info("cluster has no status yet", "Name", managedCluster.Name, "RequeueAfter:", clusterNoStatusRetryDelay)
+		return ctrl.Result{RequeueAfter: clusterNoStatusRetryDelay}, nil
+	} else if availableCondition.Status == metav1.ConditionTrue {
+		// cluster is ready
+		policies, err := r.getPolicies(ctx, managedCluster.Name)
```
imiller0, line 98: We should add a comment clarifying when policies become available. Is there a race condition here (at what point does ACM copy the policies into the child namespace)? If ACM waits until the cluster is ready to start creating child policies we could end up with a partial copy. If they are populated when the ManagedCluster is created (with the right labels) we should be able to avoid the race condition as install takes 45+ minutes.

controllers/managedclusterForCGU_controller.go
imiller0, line 136: This should be a constant.

controllers/managedclusterForCGU_controller.go
Missxiaoguo: From [ACM doc](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.4/html/governance/governance#policy-overview), it says "You can create a policy in any namespace on the hub cluster except the cluster namespace. If you create a policy in the cluster namespace, it is deleted by Red Hat Advanced Cluster Management for Kubernetes."

And I tried to create policies under cluster namespace, it was deleted like the doc describes. We should be safe.

controllers/managedclusterForCGU_controller.go
Missxiaoguo: Will do

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+//
+// Note: The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ManagedClusterForCguReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	reqLogger := r.Log.WithValues("Request.Name", req.Name)
+	reqLogger.Info("Reconciling managedCluster to create clusterGroupUpgrade")
+
+	managedCluster := &clusterv1.ManagedCluster{}
+	err := r.Get(ctx, types.NamespacedName{Name: req.Name}, managedCluster)
+	if errors.IsNotFound(err) {
```
imiller0, line 70: In a subsequent PR it would be good to handle the deletion case. If the ManagedCluster is deleted then the autogenerated CR should also be deleted.

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+const (
+	clusterNotReadyRetryDelay = time.Minute * 5
+	clusterNoStatusRetryDelay = time.Minute * 1
+	ztpInstallNS              = "ztp-install"
+	ztpDeployWaveAnnotation   = "ran.openshift.io/ztp-deploy-wave"
+)
+
+// ManagedClusterReconciler reconciles a ManagedCluster object to auto create the ClusterGroupUpgrade
+type ManagedClusterForCguReconciler struct {
```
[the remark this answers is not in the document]
Missxiaoguo: Right, I think about this as well. I don't see a reason why it's not valid. They are in different operators and using its own client and using different controller names(the one in ACM2.4 is called "managedcluster-controller").  This controller we have here is really for watching creation but not update any data.

I only test with kind cluster, I will see if I can test on a real hub cluster.

controllers/managedclusterForCGU_controller.go
Missxiaoguo: Yes yes, actually I was thinking to add the deletion if cluster is deleted. Somehow I forget to make the change. I will add it

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,207 @@
...
+func (r *ManagedClusterForCguReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	reqLogger := r.Log.WithValues("Request.Name", req.Name)
+	reqLogger.Info("Reconciling managedCluster to create clusterGroupUpgrade")
+
+	managedCluster := &clusterv1.ManagedCluster{}
+	err := r.Get(ctx, types.NamespacedName{Name: req.Name}, managedCluster)
+	if errors.IsNotFound(err) {
+		// managed cluster could have been deleted
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
Missxiaoguo: Yes

controllers/managedclusterForCGU_controller.go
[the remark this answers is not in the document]
Missxiaoguo: Tested on the hub has ACM installed. No issues.

main.go
```diff
@@ -88,6 +92,15 @@ func main() {
 	}
 	//+kubebuilder:scaffold:builder
 
+	if err = (&controllers.ManagedClusterForCguReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("ManagedClusterForCGU"),
```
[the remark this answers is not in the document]
Missxiaoguo: I think it has not been decided yet, but we can update later once we decide

go.mod
```diff
@@ -3,13 +3,17 @@ module github.com/openshift-kni/cluster-group-upgrades-operator
...
 require (
-	github.com/go-logr/logr v0.3.0
+	github.com/go-logr/logr v0.4.0
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.16.0
-	k8s.io/api v0.20.2
-	k8s.io/apimachinery v0.20.2
-	k8s.io/client-go v0.20.2
+	github.com/open-cluster-management/api v0.0.0-20210527013639-a6845f2ebcb1
+	github.com/open-cluster-management/governance-policy-propagator v0.0.0-20211209191640-b399eb8a0bf2
```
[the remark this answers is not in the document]
Missxiaoguo: I updated to use tag for open-cluster-management.io/api but cannot do that for governance-policy-propagator as it only has v2  versions, but it doesn't have v2 module

go get: github.com/open-cluster-management/governance-policy-propagator@v2.3.0-2021-06-21-09-19-16: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2

go get: github.com/open-cluster-management/governance-policy-propagator/v2@v2.3.0-2021-06-21-09-19-16: go.mod has non-.../v2 module path "github.com/open-cluster-management/governance-policy-propagator" (and .../v2/go.mod does not exist) at revision v2.3.0-2021-06-21-09-19-16

controllers/managedclusterForCGU_controller.go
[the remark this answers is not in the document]
imiller0: Excellent. Thanks for verifying!

controllers/managedclusterForCGU_controller.go
Missxiaoguo: The child policies actually get created as soon as the policy/placementrule/placementbinding are created. 

Thanks Ian for verifying/testing this on a hub cluster.  I tried it by myself as well by creating fake cluster(It will never be ready), and then creating policy/placementrule/placmentbinding. The child policy is created immediately when policy/placementrule/placmentbinding are created.

As we query the child policies after cluster is ready, this gives ACM ~45mins to create child policies, so we shouldn't have race condition here. I will update the code to remove the additional waiting.

imiller0: This means that so long as the user creates policies at approximately the same time as they create the site there should be no race condition here. :)

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,248 @@
...
+		// cluster is ready
+
+		// get child policies for the cluster
+		// child polices created as soon as placementrule returns the cluster in decision
+		// and placementrule returns decision when cluster is ready
+		policies, err := r.getPolicies(ctx, managedCluster.Name)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if len(policies.Items) == 0 {
```
imiller0, line 127: This may create a corner-case where a user-created ManagedCluster which has no policies associated with it would be requeued forever. For now I'd say just log this as a warning and don't requeue.

Missxiaoguo: Ok. Will update

controllers/managedclusterForCGU_controller.go
```diff
@@ -0,0 +1,248 @@
...
+			r.Log.Info("cluster not found", "name", req.Name)
+			clusterGroupUpgrade := &ranv1alpha1.ClusterGroupUpgrade{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      req.Name,
+					Namespace: ztpInstallNS,
+				},
+			}
+
+			r.Log.Info("Deleting clusterGroupUpgrade", "name", req.Name, "namespace", ztpInstallNS)
+			if err := r.Delete(ctx, clusterGroupUpgrade); err != nil {
```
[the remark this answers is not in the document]
Missxiaoguo: I think copied polices/placementrule/placementbinding should be cleaned up once upgrade CR deleted as we set "ownerReference" for those copied CRs.

Missxiaoguo:
In the last update, I removed the code to delete the UOCR if a managedcluster is deleted.  When creating the UOCR, the managedcluster is set as owner for this UOCR(It will contain a ownerReferences section), so if the cluster is deleted, the UOCR will get cleaned up automatically. And I verified this on a hub cluster as well. 

"ownerReferences" is the common usecase for the resources created by a operator.

imiller0:
> In the last update, I removed the code to delete the UOCR if a managedcluster is deleted. When creating the UOCR, the managedcluster is set as owner for this UOCR(It will contain a ownerReferences section), so if the cluster is deleted, the UOCR will get cleaned up automatically. And I verified this on a hub cluster as well.

Can you add a comment in the code (near line 74) where the ManagedCluster is not found which describes how the deletion case works. This will be very helpful in maintenance.

Missxiaoguo:
> > In the last update, I removed the code to delete the UOCR if a managedcluster is deleted. When creating the UOCR, the managedcluster is set as owner for this UOCR(It will contain a ownerReferences section), so if the cluster is deleted, the UOCR will get cleaned up automatically. And I verified this on a hub cluster as well.
> 
> Can you add a comment in the code (near line 74) where the ManagedCluster is not found which describes how the deletion case works. This will be very helpful in maintenance.

I described the case in Line 63-64 and L186-187. I think it's enough?

imiller0:
> > > In the last update, I removed the code to delete the UOCR if a managedcluster is deleted. When creating the UOCR, the managedcluster is set as owner for this UOCR(It will contain a ownerReferences section), so if the cluster is deleted, the UOCR will get cleaned up automatically. And I verified this on a hub cluster as well.
> > 
> > 
> > Can you add a comment in the code (near line 74) where the ManagedCluster is not found which describes how the deletion case works. This will be very helpful in maintenance.
> 
> I described the case in Line 63-64 and L186-187. I think it's enough?

Yes, I see those now. Thank you

imiller0 (approved):

PR #48, opened 2021-12-15
Add basic blocking mechanisms solution for multiple CRs

irinamihai:
Description:
- when an upgrade becomes enable, check if there are other upgrade CRs currently running (with status upgradeNotCompleted) and if there are, don't start the current upgrade until they are done
- add new member in ClusterGroupUpgradeSpec: BlockingCRs to hold a list of both the blocking CRs and their namespaces
- the current basic functionality assumes that if UOCR A contains B as blocking CR, then B should also contain A as blocking
- add KUTTL tests for this scenario


config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
```diff
@@ -39,6 +39,17 @@ spec:
           spec:
             description: ClusterGroupUpgradeSpec defines the desired state of ClusterGroupUpgrade
             properties:
+              blockingCRs:
+                items:
+                  description: BlockingCR defines the Upgrade CRs that block the current
```
browsell, line 44: nit. maybe reword. 
BlockingCR defines the running Upgrade CRs that block the current CR from running

api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -33,6 +33,12 @@ type RemediationStrategySpec struct {
 	Timeout int `json:"timeout,omitempty"`
 }
 
+// BlockingCR defines the Upgrade CRs that block the current CR from running if they are currently running
```
browsell, line 36: nit. maybe reword. 
BlockingCR defines the running Upgrade CRs that block the current CR from running

config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
serngawy: I believe the blockingCRs must go under status not spec otherwise the ArgoCD/ACM App that sync the ClusterGroupUpgrade.yaml from the git repo will detect there are different and it may re-override it.

serngawy (commented):
Just to clarify the logic , why this is an issue to block ? If I have CGU that update sriov operator for cluster x AND i have another CGU that update ptp operator for cluster z. Why do we need to block that ? Do I miss understand

config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
irinamihai: @serngawy , the blocking CRs should be under spec as we are specifying which UOCRs block the current CR from running, it's not a status that we are updating. If we want to use the blocking feature, then they should also be present in the ClusterGroupUpgrade.yaml from the git repo. They are optional though.

serngawy: okay, if it will be defined in the ClusterGroupUpgrade before adding it to the hub that will work.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1241,6 +1266,31 @@ func (r *ClusterGroupUpgradeReconciler) updateStatus(ctx context.Context, cluste
...
+		err := r.Get(ctx, types.NamespacedName{Name: blockingCR.Name, Namespace: blockingCR.Namespace}, cgu)
+
+		if err != nil {
+			r.Log.Info("[blockingCRsRunning] CR not found", "name", blockingCR.Name)
+			return nil, err
+		}
+
+		r.Log.Info("[blockingCRsRunning] condition ", "cgu.Status.Conditions", cgu.Status.Conditions)
+		for i := range cgu.Status.Conditions {
+			if cgu.Status.Conditions[i].Reason == "UpgradeNotCompleted" {
```
imiller0, line 1285: Does this status exist on an upgrade CR as soon as it is created? If so, this is OK. If this status is only added once the upgrade starts I think the logic needs to be changed to check for any blocking CR that is not completed. I am worried that the user creates CR "y" which is blocked by "x". If "x" has not started yet for some reason and so it does not have "UpgradeNotCompleted" then "y" will start running even though "x" hasn't completed.

controllers/clustergroupupgrade_controller.go
```diff
@@ -113,13 +114,37 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
-					clusterGroupUpgrade.Status.Status.StartedAt = metav1.Now()
+				requeueAfter := 5 * time.Minute
+
+				// If the remediation plan is empty, update the status.
+				if clusterGroupUpgrade.Status.RemediationPlan == nil {
+					statusReason = "UpgradeCannotStart"
+					statusMessage = "The ClusterGroupUpgrade CR has no NonCompliant clusters for the specified managed policies"
+				} else if clusterGroupUpgrade.Spec.Enable == true {
+					// Check if there are any CRs running that are blocking the start of the current one.
+					blockingCRs, err := r.blockingCRsRunning(ctx, clusterGroupUpgrade)
```
imiller0, line 129: Minor nit on naming. A blocking CR that is pending (not yet running) should still block this CR. See note in blockingCRsRunning below.

controllers/clustergroupupgrade_controller.go
irinamihai: Well, in that scenario then "y" starts and has its status "UpgradeNotCompleted", then when "x" is ready to start, it won't because it's blocked by "y"'s status. This is, of course, if both "x" and "y" define each other as blocking. This feature ensures "x" and "y" don't run at the same time, but it doesn't mean "x" should run first.

PR #49, opened 2021-12-15
Update name of the ConfigurationPolicy from copied policies

irinamihai:
Description:
- if the managed policies and the copied policies share the same ConfigurationPolicy in their policy-templates, then we are
  getting an error from ACM that the names must be different
- update the ConfigurationPolicy for the copied policies to also include the name of the UOCR

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -572,7 +572,7 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyNameForCopiedPo
 		}
 		// Update the metadata name
 		metadataContent := metadata.(map[string]interface{})
-		metadataContent["name"] = clusterGroupUpgrade.Name + "-" + metadataContent["name"].(string)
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -544,19 +549,47 @@ func (r *ClusterGroupUpgradeReconciler) copyManagedInformPolicy(
...
+			return fmt.Errorf("Policy %s is missing its spec.policy-templates.objectDefinition", managedPolicyName)
+		}
+		objectDefinitionContent := objectDefinition.(map[string]interface{})
+		metadata := objectDefinitionContent["metadata"]
+		if metadata == nil {
+			return fmt.Errorf("Policy %s is missing its spec.policy-templates.objectDefinition.metadata", managedPolicyName)
+		}
+		// Update the metadata name
+		metadataContent := metadata.(map[string]interface{})
+		metadataContent["name"] = clusterGroupUpgrade.Name + "-" + metadataContent["name"].(string)
```
imiller0, line 575: This should call getResourceName for consistent handling of "prefixing" names.

PR #58, opened 2021-12-20
Split dockerfile to enable CI

vitus133:
Specifying a target is not supported on OpenShift Builds
that ci-operator is using to build images.
(https://docs.openshift.com/container-platform/4.9/rest_api/workloads_apis/build-build-openshift-io-v1.html)
/cc @browsell @irinamihai

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,13 +162,13 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 docker-build: ## Build docker image with the manager.
-	docker build -t ${IMG} --target operator_image .
+	docker build -t ${IMG} -f Dockerfile .
 
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
 
 docker-build-precache: ## Build pre-cache workload docker image.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -162,13 +162,13 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 docker-build: ## Build docker image with the manager.
-	docker build -t ${IMG} --target operator_image .
+	docker build -t ${IMG} -f Dockerfile
```
irinamihai, line 165: I think we need the last dot at the end.

Makefile
```diff
@@ -162,13 +162,13 @@ run: manifests generate fmt vet ## Run a controller from your host.
...
 
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
 
 docker-build-precache: ## Build pre-cache workload docker image.
-	docker build -t ${PRECACHE_IMG} --target precache_image .
+	docker build -t ${PRECACHE_IMG} -f Dockerfile.precache
```
irinamihai, line 171: Same comment as above.

PR #62, opened 2021-12-22
Fix the wave sorting issue

Missxiaoguo:
Deploy wave is a string type. Convert it to int before sorting.

### DIFF
--- a/controllers/managedclusterForCGU_controller.go
+++ b/controllers/managedclusterForCGU_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


controllers/managedclusterForCGU_controller.go
```diff
@@ -156,8 +157,13 @@ func (r *ManagedClusterForCguReconciler) newClusterGroupUpgrade(ctx context.Cont
 	for _, policy := range policies.Items {
 		deployWave, found := policy.GetAnnotations()[ztpDeployWaveAnnotation]
 		if found {
+			deployWaveInt, err := strconv.Atoi(deployWave)
+			if err != nil {
+				// err convert from string to int
+				return err
```
imiller0, line 163: It would be helpful to return an error which indicates that in policy xxxxxx the ran.openshift.io/ztp-wave-deploy value is not an integer.

Missxiaoguo: Done

imiller0 (approved):
/lgtm

PR #64, opened 2021-12-23
Precaching

vitus133:
This is a pre-caching code ready for review and integration.
It also includes https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/63
The logic of getting the right upgrade spec from the right policies is still being refined
/cc @irinamihai @browsell @Missxiaoguo


controllers/precache.go
```diff
@@ -0,0 +1,768 @@
+package controllers
```
vitus133, line 1: This is a new large file - not shown by github by default, hence adding this comment

vitus133:
Too much action is still happening here. Will open a new one when ready

PR #65, opened 2022-01-03
Add precaching

vitus133:
Container image pre-caching integration
/cc @browsell @Missxiaoguo @imiller0 @nishant-parekh 
/assign @irinamihai


config/pre-cache/overrides.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
browsell, line 1: What would the workflow be here. 
Admin applies the override ConfigMap then applies a upgrade cr ?

vitus133: Yes

PR #69, opened 2022-01-20
Support actions before/after upgrade

Missxiaoguo:
This commit included the following changes:
  - Update CRD to support adding/deleting cluster labels before
    enable upgrade and after upgrade is completed
  - Move deleteResources action under "spec.actions.afterCompletion"
  - Update managedclusterForCGU controller to add proper label actions
    in generated UOCR
  - Prevent managedClusterForCGU controller to generate UOCR when
    ztp-done label applied to the cluster
  - Monitor delete action for owned UOCR so new UOCR can be automatically
    created when the old one is deleted

This commit fixs Bug2038253 and Bug2040874


Missxiaoguo:
@irinamihai @imiller0 @vitus133 @browsell

controllers/actions.go
```diff
@@ -0,0 +1,160 @@
...
+				continue
+			}
+			return err
+		}
+		if err := r.addClusterLabels(ctx, managedCluster, labels["add"]); err != nil {
+			return fmt.Errorf("Fail to add labels for cluster %s: %v", managedCluster.Name, err)
+		}
+		if err := r.deleteClusterLabels(ctx, managedCluster, labels["delete"]); err != nil {
+			return fmt.Errorf("Fail to delete labels for cluster %s: %v", managedCluster.Name, err)
+		}
```
imiller0, line 89: Does this generate an error if the label does not exist?  If this fails because the cluster does not have the specified label then we should not error out. The intent of the "deleteLables" is that the cluster should not have the label, which is the case if it does not exist.

Missxiaoguo: No error for the non-existence label. Delete is a no-op if label doesn't exist. I added two more unittests to cover this.

controllers/managedclusterForCGU_controller.go
```diff
@@ -47,6 +47,8 @@ const (
 	clusterStatusCheckRetryDelay = time.Minute * 5
 	ztpInstallNS                 = "ztp-install"
 	ztpDeployWaveAnnotation      = "ran.openshift.io/ztp-deploy-wave"
+	ztpIsRunningLabel            = "ztp-is-running"
```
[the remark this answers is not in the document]
Missxiaoguo: Sure, I have updated

controllers/managedclusterForCGU_controller.go
```diff
@@ -239,12 +266,20 @@ func (r *ManagedClusterForCguReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
 				CreateFunc:  func(e event.CreateEvent) bool { return true },
 				DeleteFunc:  func(e event.DeleteEvent) bool { return false },
 				UpdateFunc:  func(e event.UpdateEvent) bool { return false },
 			})).
+		Owns(&ranv1alpha1.ClusterGroupUpgrade{},
+			// watch for delete event for owned ClusterGroupUpgrade
+			builder.WithPredicates(predicate.Funcs{
+				GenericFunc: func(e event.GenericEvent) bool { return false },
+				CreateFunc:  func(e event.CreateEvent) bool { return false },
+				DeleteFunc:  func(e event.DeleteEvent) bool { return true },
```
imiller0, line 281: With monitoring for the delete is the behavior that we will be enforcing that a CGU will be created for every ManagedCluster until that ManagedCluster has the ztp-done label attached? If so we need to clearly document that for the user. I can see a user getting confused when the delete a CGU and it immediately comes back. 
I would also update the info message on line 248 to read "Found ManagedCluster <managedClusterName> without ztp-done label. Created clusterGroupUpgrade <name>"

config/crd/bases/ran.openshift.io_clustergroupupgrades.yaml
```diff
@@ -39,6 +39,58 @@ spec:
...
+                        description: This field defines a map of key/value pairs that
+                          identify the cluster labels to be deleted for the specified
+                          clusters. Labels applied to the clusters either defined
+                          in spec.clusters or selected by spec.clusterSelector.
+                        type: object
+                      deleteObjects:
+                        default: true
+                        description: This field defines whether clean up the resources
+                          created for upgrade
+                        type: boolean
```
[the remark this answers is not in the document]
Missxiaoguo: What I can think of is to re-use the copied policies/placementrules/placementbindings for next UOCR and keep the history for user to check what has been done for this upgrade.

controllers/managedclusterForCGU_controller.go
imiller0: In general I think it is better to have the user take a direct action that says "I want to restart this". Having a restart be triggered by deleting a CR is a bit non-intuitive. However, including this watch on the delete does have the benefit that the behavior of the controller when restarted matches the running behavior. For example, if a CGU get's deleted this new logic will re-create the CGU. The same would happen if the controller restarted for some reason. Both behaviors are now aligned (which is good). So, I'm ok with leaving it but I think we should add a story to allow the user to affirmatively say that they want to restart the CGU (eg disable and then re-enable the CGU).

imiller0 (approved):
/lgtm

PR #70, opened 2022-01-24
Include the possibility to use podman and docker as container engine

alosadagrande:
A few changes in the Makefile so you can choose to use either podman or docker to build and deploy the CGU operator. By default the behaviour is still using docker as the container engine, however by passing the build arg ENGINE=podman changes the behaviour.

A few changes in the comments as well, since IMO there is no such thing as docker images. 

I tested make docker-build, docker-push and docker-deploy options with ENGINE=podman successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.0.3
 
+# You can use podman or docker as a container engine. Notice that there are some options that might be only valid for one of them.
+ENGINE ?= docker
+
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "preview,fast,stable")
 # To re-generate a bundle for other specific channels without changing the standard setup, you can:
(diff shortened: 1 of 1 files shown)


bundle/manifests/cluster-group-upgrades-operator.clusterserviceversion.yaml
```diff
@@ -170,7 +170,7 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
-                image: quay.io/openshift-kni/cluster-group-upgrades-operator:v0.0.3
+                image: quay.io/alosadag/cluster-group-upgrades-operator:latest
```
Missxiaoguo, line 173: Looks like you forgot to change it back to openshift-kni, same as below.
Also I think the image tag should still be v0.0.3, in the Makefile,  we have 
VERSION ?= 0.0.3
BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)

alosadagrande: Fixed, thank you @Missxiaoguo

PR #72, opened 2022-01-26
[BZ#2034625] Approve InstallPlan for Subscription type policies

irinamihai:
Description:
- parse managed policies to determine if they contain a Subscription object
- for policies that contain a Subscription, create a ManagedClusterView
  to get access to the details about the InstallPlan
- create a ManagedClusterView to get the latest resource version for the
  InstallPlan
- create a ManagedClusterAction to approve the InstallPlan only if the
  approval is set to Manual
- at the end of the upgrade, delete the created multicloud objects
- have all the functions dealing with multicloud operator objects in a
  separate file: multicloud_util.go
- TODO: integration tests


vitus133 (commented):
Hi, great work! Added some comments, but most of them can be addressed at a later point, if the code freeze is a factor here.
There is also a couple of utility functions that we'll need to de-duplicate one day (for example stripping the objects out of policies).
Cheers,
VG

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,330 @@
...
+			},
+		}
+
+		mcv.Spec = viewSpec
+		if err := c.Update(ctx, mcv); err != nil {
+			return nil, err
+		}
+	}
+	// Give some time for the ManagedClusterView to retrieve the info.
+	time.Sleep(ManagedClusterActionWaitTimeSec * time.Second)
```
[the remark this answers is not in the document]
irinamihai: Yes, I agree, however after consulting with ACM folks, it felt like the better option. It's either we wait a couple of seconds now and have a really good chance of getting the info we need or wait 5 minutes until the next reconcile loop. It can save quite a lot of time in the end doing it this way.

> Irina Mihai Jan 20th at 16:42
> Hi all! What's the average time we should expect a managedClusterAction CR to be Completed? I've noticed it's usually pretty fast, but was wondering what's the amount of time we should wait before checking its status.
> Thank you!
> 2 replies
> Dangpeng Liu:no_entry:  7 days ago
> Actually, there is no average time for that. We just watch the managedClusterAction CR and apply the resource in the managedClusterAction template. So the response time may be affected by managedCluster api server.
> A better way is check the status condition Completed periodically (interval could be 1-3 seconds). And set an timeout (may be 1-2 minute) (edited)

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func EnsureManagedClusterActionForInstallPlan(
+	ctx context.Context, c client.Client, name string, namespace string,
+	installPlan operatorsv1alpha1.InstallPlan, Log logr.Logger) (*actionv1beta1.ManagedClusterAction, error) {
+
+	mcaForInstallPlan := &actionv1beta1.ManagedClusterAction{}
+	if err := c.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, mcaForInstallPlan); err != nil {
+		// If the specific managedClusterAction was not found, create it.
+		if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
irinamihai: Yes, the operator is creating the resources, but you never know what happens on a system, so better be safe and add as many checks.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	return actionSpec, nil
+}
+
+func CheckManagedClusterActionDone(
+	ctx context.Context, c client.Client, mcaName string, mcaNamespace string, Log logr.Logger) error {
+	// Check the managedClusterAction has completed successfully.
+	mca := &actionv1beta1.ManagedClusterAction{}
+	err := c.Get(ctx, types.NamespacedName{Name: mcaName, Namespace: mcaNamespace}, mca)
```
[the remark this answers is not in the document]
irinamihai: It does work if you query the action within 60 seconds. It gets deleted after that if it has completed.

controllers/clustergroupupgrade_controller.go
```diff
@@ -374,30 +378,89 @@ func (r *ClusterGroupUpgradeReconciler) getNextRemediationPoliciesForBatch(
...
 }
 
+func (r *ClusterGroupUpgradeReconciler) approveInstallPlan(
+	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade,
+	managedPolicyIndex int, clusterName string) error {
+	managedPolicyName := clusterGroupUpgrade.Spec.ManagedPolicies[managedPolicyIndex]
+
+	// If there is no content saved for the current managed policy, return.
+	_, ok := clusterGroupUpgrade.Status.ManagedPoliciesContent[managedPolicyName]
+	if ok == false {
```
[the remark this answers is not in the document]
irinamihai: We can keep this in mind for the future and maybe add extra checks as part of the CI job.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1023,6 +1189,78 @@ func (r *ClusterGroupUpgradeReconciler) reconcileResources(ctx context.Context,
 	return nil
 }
 
+func (r *ClusterGroupUpgradeReconciler) getPolicyClusterStatus(policy *unstructured.Unstructured) ([]interface{}, error) {
+	policyName := policy.GetName()
+
+	// Get the compliant status part of the policy.
+	if policy.Object["status"] == nil {
+		return nil, fmt.Errorf("Policy %s is missing its status", policyName)
```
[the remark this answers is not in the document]
irinamihai: A policy could be missing its status when it doesn't have a placement rule or up until the field is populated once the policy was matched with a placement rule. At the point this function is run, if a policy doesn't have its status then it means something pretty bad happened to the system, considering we had the 

I think it would be a good idea to treat some of these errors with temporary conditions not just here, but throughout the code with most of the ACM resources. I would rather put this on the list with other future improvements we want to do.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return mcaForInstallPlan, nil
+}
+
+// NewManagedClusterActionForInstallPlanSpec returns the action spec for approving an InstallPlan.
+func NewManagedClusterActionForInstallPlanSpec(installPlan operatorsv1alpha1.InstallPlan) (actionv1beta1.ActionSpec, error) {
+	installPlanClusterServiceVersionNames, err := json.Marshal(installPlan.Spec.ClusterServiceVersionNames)
+	if err != nil {
+		fmt.Printf("Error trying to Marshal InstallPlan ClusterServiceVersionNames: %s", err.Error())
```
Missxiaoguo, line 256: Looks like we are missing a return here.

api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -124,18 +124,26 @@ type PrecachingStatus struct {
...
-	ManagedPoliciesNs map[string]string  `json:"managedPoliciesNs,omitempty"`
-	Status            UpgradeStatus      `json:"status,omitempty"`
-	Precaching        PrecachingStatus   `json:"precaching,omitempty"`
+	PlacementBindings      []string           `json:"placementBindings,omitempty"`
+	PlacementRules         []string           `json:"placementRules,omitempty"`
+	CopiedPolicies         []string           `json:"copiedPolicies,omitempty"`
+	Conditions             []metav1.Condition `json:"conditions,omitempty"`
+	RemediationPlan        [][]string         `json:"remediationPlan,omitempty"`
+	ManagedPoliciesNs      map[string]string  `json:"managedPoliciesNs,omitempty"`
+	ManagedPoliciesContent map[string]string  `json:"managedPoliciesContent,omitempty"`
```
Missxiaoguo, line 144: If we plan to have one entry for managedpolicies information(name, namespace, complaint status), I think this can be moved to the structure as well. Not ugrent definitely can be a future enhancement :)

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+
+	if err != nil {
+		// If the specific managedClusterView was not found, create it.
+		if errors.IsNotFound(err) {
+			Log.Info("[EnsureManagedClusterView] MCV doesn't exist, create it", "name", name, "namespace", namespace)
+			viewMeta := metav1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+				Labels: map[string]string{
+					"openshift-cluster-group-upgrades/clusterGroupUpgrade": cguLabel,
```
Missxiaoguo, line 156: nit: "openshift-cluster-group-upgrades/clusterGroupUpgrade"is being frequently used. Consider defining a const var.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+			mcaForInstallPlan = &actionv1beta1.ManagedClusterAction{
+				ObjectMeta: actionMeta,
+				Spec:       actionSpec,
+			}
+
+			if err := c.Create(ctx, mcaForInstallPlan); err != nil {
+				return mcaForInstallPlan, nil
+			}
+		}
```
Missxiaoguo, line 233: For errors other than IsNotFound, I think we want to return err?

controllers/utils/multicloud_util.go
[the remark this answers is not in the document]
imiller0: This is a case where an exponential backoff loop for the reconcile time would be helpful. On the first pass the loop would re-trigger after 5 seconds, then 10, then 20, up to the max limit.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// GetMultiCloudObjectName computes the name of a view or action
+func GetMultiCloudObjectName(managedClusterType string, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, kind string, objectName string) string {
+	return strings.ToLower(managedClusterType + "-" + clusterGroupUpgrade.Name + "-" + kind + "-" + objectName)
```
imiller0, line 340: If there are two CGU with the same name in different namespaces they could generate the same named MCA or MCV. If you add the CGU namespace then that small case is avoided.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+func NewManagedClusterActionForInstallPlanSpec(installPlan operatorsv1alpha1.InstallPlan) (actionv1beta1.ActionSpec, error) {
+	installPlanClusterServiceVersionNames, err := json.Marshal(installPlan.Spec.ClusterServiceVersionNames)
+	if err != nil {
+		fmt.Printf("Error trying to Marshal InstallPlan ClusterServiceVersionNames: %s", err.Error())
+	}
+
+	templateContent := fmt.Sprintf(
+		`{"apiVersion": "operators.coreos.com/v1alpha1","kind": "InstallPlan",
+		  "metadata": {"name": "%s","resourceVersion": "%s"},
+		  "spec": {"approval": "Manual","approved": true, "clusterServiceVersionNames": %s}}`,
```
imiller0, line 262: This should get applied as a patch in which case I don't think we want to include the "approval: Manual" field (unless we have to). Including it means that any other logic error in the code could lead to us changing the approval type of an installplan (obviously without meaning to). Same thing for the CSV version, we would not want to overwrite/update that, it should already be set to the right value.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+		mcaName := GetMultiCloudObjectName(
+			ManagedClusterActionPrefix, clusterGroupUpgrade, "InstallPlan", installPlan.Name)
+		_, err := EnsureManagedClusterActionForInstallPlan(ctx, c, mcaName, clusterName, installPlan, Log)
+		if err != nil {
+			return InstallPlanCannotBeApproved, err
+		}
+
+		// Check that the ManagedClusterAction completed successfully. If so, then there is no need to also check
+		// the install plan.
+		time.Sleep(ManagedClusterActionWaitTimeSec * time.Second)
```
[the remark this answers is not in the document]
imiller0: There is a comment below which says the ManagedClusterAction gets deleted 60 seconds after it completes. If the reconcile loop for this controller is 5 minutes then will we miss reading the status from any pass through here in which the action takes 4+ seconds to complete?

controllers/utils/multicloud_util.go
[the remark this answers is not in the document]
imiller0: Not sure how feasible this is but we should ask if the ManagedClusterAction api can be updated to allow us to register for a status update. When the MCA status changes (done, error, etc) it would copy the status to our CR before deleting itself.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// DeleteMultiClusterObjects cleans up views associated to an UOCR.
+func DeleteMultiClusterObjects(
+	ctx context.Context, c client.Client, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, Log logr.Logger) error {
+	// Only need to delete ManagedClusterView as ManagedClusterActions get deleted automatically.
+	for _, batch := range clusterGroupUpgrade.Status.RemediationPlan {
+		for _, clusterName := range batch {
+			var mcvLabels = map[string]string{
+				"openshift-cluster-group-upgrades/clusterGroupUpgrade": clusterGroupUpgrade.Name}
```
imiller0, line 316: If we have two CGU in different namespaces working with the same cluster then we could end up with one of the CGU deleting the multicluster objects from the other. If we add the CGU namespace to the label we would avoid any chance this happens.

deploy/acm/crds/action.open-cluster-management.io_managedclusteractions.yaml
```diff
@@ -0,0 +1,121 @@
+
```
imiller0, line 1: Is this included just for testing? How does the operator ensure this CRD is available when installed via subscription?

deploy/acm/crds/action.open-cluster-management.io_managedclusteractions.yaml
```diff
@@ -0,0 +1,121 @@
...
+---
```
imiller0, line 2: No need to include separator at the top of the file. Technically it makes it appear that there are 2 sections here, one empty and one with the CRD.

deploy/acm/crds/view.open-cluster-management.io_managedclusterviews.yaml
```diff
@@ -0,0 +1,126 @@
+
+---
```
imiller0, line 2: Same comments as previous CRD

deploy/acm/policies/patch-policies-status.sh
```diff
@@ -20,3 +20,8 @@ curl -k -s -X PATCH -H "Accept: application/json, */*" \
 http://localhost:8001/apis/policy.open-cluster-management.io/v1/namespaces/$4/policies/policy4-common-sriov-sub-policy/status \
 --data '{"status":{"compliant":"NonCompliant","status":[{"clustername":"spoke4","clusternamespace":"spoke4","compliant":"NonCompliant"}, {"clustername":"spoke5","clusternamespace":"spoke5","compliant":"NonCompliant"}, {"clustername":"spoke6","clusternamespace":"spoke6","compliant":"NonCompliant"}]}}'
 
+curl -k -s -X PATCH -H "Accept: application/json, */*" \
+-H "Content-Type: application/merge-patch+json" \
+http://localhost:8001/apis/policy.open-cluster-management.io/v1/namespaces/$4/policies/policy5-subscriptions/status \
+--data '{"status":{"compliant":"NonCompliant","status":[{"clustername":"spoke4","clusternamespace":"spoke4","compliant":"NonCompliant"}, {"clustername":"spoke5","clusternamespace":"spoke5","compliant":"NonCompliant"}, {"clustername":"spoke6","clusternamespace":"spoke6","compliant":"NonCompliant"}]}}'
+
```
imiller0, line 27: Is this test code, or used in production?

controllers/utils/multicloud_util.go
irinamihai: I think this is the case with a couple of other labels. Will add it to the enhancements list.

controllers/utils/multicloud_util.go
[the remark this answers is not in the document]
irinamihai: I think as you've suggested below, a better approach would be to do incremental wait of 5 seconds, 10, 20 up to 1 minute. I will check with ACM if we can configure to either not have the action deleted or to configure when it gets deleted.

deploy/acm/crds/action.open-cluster-management.io_managedclusteractions.yaml
irinamihai: This is just for testing. If ACM is installed then the CRD should exist. I didn't have to do anything extra on my environment to get these CRDs.

deploy/acm/policies/patch-policies-status.sh
irinamihai: Just for testing.

controllers/utils/multicloud_util.go
irinamihai: This is not how actions work, unfortunately. Even though I set the action type to update: "ActionType: actionv1beta1.UpdateActionType" (in NewManagedClusterActionForInstallPlanSpec), these fields are still required, otherwise you can't apply them, including the CSV name - this is the reason I had to create the managedClusterView for the installPlan, to get access to the CSV and resource version. Also, that is why I made the checks above to make sure the installPlan has the approval set to Manual.

@serngawy had the same surprise when he discovered we have to include the resource version. He can comment/conform/share more options and I will triple check once I can recover my setup.

imiller0: The resourceVersion I kind of understand (helps to uniquely identify the target of the action perhaps?), but the requirement to include fields that we don't intend (or want) to modify is strange. We should raise this with the ACM team to understand why we can't patch just one field without specifying the others.

controllers/utils/multicloud_util.go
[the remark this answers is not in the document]
irinamihai: I haven't received an answer from ACM, but from a previous suggestion from them and after talking to @serngawy, incremental checks seem like a good option.

controllers/utils/multicloud_util.go
irinamihai: I did followup on a discussion Mohamed had with ACM on this same topic. Their answer to Mohamed's question "why do we also need resourceVersion" was "Currently, I think it’s necessary for some resources. But I think we could investigate if it’s possible to  enhance it in our code", so it doesn't seem the same for all the resources. We'll see what they say, but for now we need all the fields in order for this to work.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,360 @@
...
+	ctx context.Context, c client.Client, mcaName string, mcaNamespace string, timeout int) error {
+	// Check the managedClusterAction has completed successfully.
+	mca := &actionv1beta1.ManagedClusterAction{}
+
+	// Do incremental checks until we can decide if the ManagedClusterAction was successful or not.
+	elapsedTimeSec := 0
+	waitTimeSec := ManagedClusterActionWaitTimeSec
+	runCount := 1
+
+	for elapsedTimeSec <= timeout {
```
imiller0, line 296: In a heavily loaded system the amount of time it takes to make API calls be be several seconds. This can really skew the logic of waiting based on adding the sleep time. In this case maybe up to ~25 additional seconds which may be OK but we should look to improve in a follow-on.

controllers/clustergroupupgrade_controller.go
```diff
@@ -253,10 +256,15 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 			} else {
 				// Add the needed cluster names to upgrade to the appropriate placement rule.
-				err = r.addClustersToPlacementRule(ctx, clusterGroupUpgrade)
+				reconcileSooner, err := r.remediateCurrentBatch(ctx, clusterGroupUpgrade)
 				if err != nil {
 					return ctrl.Result{}, err
 				}
+				if reconcileSooner == true {
+					requeueAfter := 30 * time.Second
```
imiller0, line 264: Better to have the 30 as a constant for "fast requeue", but this can be done in a subsequent cleanup.

controllers/utils/multicloud_util.go
```diff
@@ -0,0 +1,306 @@
...
+	// We do this by checking it's status and reason.
+	if conditionMCVforSub.Status == "True" && conditionMCVforSub.Reason == viewv1beta1.ReasonGetResource {
+		// Get the subscription content from the ManagedClusterView.
+		subscription := operatorsv1alpha1.Subscription{}
+		json.Unmarshal([]byte(mcv.Status.Result.Raw), &subscription)
+
+		// If the subscription's status is "UpgradePending" approve the installPlan. For any other value of the state, continue.
+		if subscription.Status.State != SubscriptionStateUpgradePending {
+			multiCloudLog.Info("Subscription State is not pending upgrade",
+				"subscription", subscription.ObjectMeta.Name, "namespace", subscription.ObjectMeta.Namespace)
```
imiller0, line 55: In a subsequent cleanup pass, this message may be slightly confusing to a user in the case where a subscription is included in the managedPolicies list but is already in the "AtLatestKnown" state. In that case we should either print nothing (since there is nothing for us to do) or change the Info log text to something like "Subscription is already at latest known".

imiller0 (commented):
/lgtm

controllers/clustergroupupgrade_controller.go
irinamihai: Yes, there are a couple of requeue values that should be made constants.

PR #74, opened 2022-01-27
Update precache image tag in csv

vitus133:
Automatically patch pre-cache image tag in csv
to the operator version.

/cc @imiller0 @nishant-parekh @irinamihai

### DIFF
--- a/hack/patch-precache-image.py
+++ b/hack/patch-precache-image.py
@@ -1,0 +1,14 @@
+#!/usr/bin/python3
+import yaml
+import sys
+
+file = sys.argv[1]
+pull = sys.argv[2]
+with open(file,'r') as f:
+    obj = yaml.safe_load(f)
+for item in obj.get('spec',{}).get('relatedImages',{}):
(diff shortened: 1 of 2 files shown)


hack/patch-precache-image.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/python3
+import yaml
+import sys
+
+file = sys.argv[1]
+pull = sys.argv[2]
+with open(file,'r') as f:
+    obj = yaml.safe_load(f)
+for item in obj.get('spec',{}).get('relatedImages',{}):
```
rcarrillocruz, line 9: is it assured there will always be a relatedImages field?

vitus133: Yes, because of this:
https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/cb2f0bcd5f0b61bd686603485f2d38c7a793ae55/config/manifests/pre-cache-image-patch.yaml#L7

PR #81, opened 2022-02-06
Add universal index processing

vitus133:
OLM catalogs are switching from sqlite-based to file-system based
index format (also referenced as FBC). The enhancement is described in
https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md
Additional information is available here:
https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/512
The switch of default Red Hat catalogs to FBC is expected to occur in Openshift 4.11.
This commit changes the way catalog processing is
done by pre-caching workload to extract 
packages-bundles-channels relations from the
index image. It is switching to a new opm command
(`opm render`), that works for both sqlite-based catalogs and FBC. The command is supported
by opm included in catalogs starting from 4.9 onwards

/cc @browsell @irinamihai @imiller0 @rcarrillocruz @danielmellado

### DIFF
--- a/pre-cache/olm
+++ b/pre-cache/olm
@@ -4,18 +4,18 @@ cwd="${cwd:-/opt/precache}"
 . $cwd/common
 
 podman_pull_path="${podman_pull_path:-/run/containers/0}"
-index_download_path="${index_download_path:-/cache/downloaded}"
+rendered_index_path="${rendered_index_path:-/tmp/index.json}"
 
 
 extract_pull_spec(){
-    index_download_path=$1
+    rendered_index=$1
     operators_spec_file=$2
     pull_spec_file=$3
-    /host/usr/libexec/platform-python /opt/precache/parse_index.py "${index_download_path}" "${operators_spec_file}" "${pull_spec_file}"
+    /host/usr/libexec/platform-python /opt/precache/parse_index.py "${rendered_index}" "${operators_spec_file}" "${pull_spec_file}"
     return $?
 }
 
-export_index(){
+render_index(){
     index=$1
     pull_secret_path=$2
     packages=$3
@@ -35,10 +35,12 @@ export_index(){
     else
         opm_bin=$image_mount/bin/opm
     fi
-    $opm_bin index export -i $index -p $packages -c podman -f $index_download_path
+
+    $opm_bin render $index > $rendered_index_path
     if [[ $? -ne 0 ]]; then
         return 1
     fi
+
 }
 
 extract_packages(){
@@ -53,7 +55,6 @@ extract_packages(){
 }
 
 olm_main(){
-    declare -i ret_val=0
     if ! [[ -n $(sort -u $config_volume_path/operators.indexes) ]]; then
       log_debug "Operators index is not specified. Operators won't be pre-cached"
       return 0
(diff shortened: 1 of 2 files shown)


pre-cache/parse_index.py
```diff
@@ -27,54 +24,46 @@ def parse_args():
...
-            bundle_dirs, key=lambda dir_name: SequenceMatcher(
-                None, csv_name, dir_name).ratio())[-1]
-        bundles.append(os.path.join(
-            args.index_download_path, pkg_name, bundle_dir))
-    return bundles
-
-
-def extract_images(bundles):
+        channels[item[0].strip()] = item[1].strip()
+        packages.append(item[0].strip())
```
rcarrillocruz, line 35: Maybe save item[0].strip() in a variable, since you are referencing them twice consecutively.

pre-cache/parse_index.py
```diff
@@ -27,54 +24,46 @@ def parse_args():
...
-        bundle_dir = sorted(
-            bundle_dirs, key=lambda dir_name: SequenceMatcher(
-                None, csv_name, dir_name).ratio())[-1]
-        bundles.append(os.path.join(
-            args.index_download_path, pkg_name, bundle_dir))
-    return bundles
-
-
-def extract_images(bundles):
+        channels[item[0].strip()] = item[1].strip()
```
rcarrillocruz, line 34: What is stored in item[0] and item[1] ?

danielmellado: Also, is there a possibility for item[0] or item[1] to be empty? Checking the code from above it seems that you're reading a file and these would be strings, but I'd check that case I mentioned.

vitus133:
/hold 
@rcarrillocruz @danielmellado Thanks for your comments!
I'm putting this on hold for now, since I don't want it to land in 4.10
Will address your comments next week, together with one more small issue I found (Need to configure the opm with pull secret in .docker/config.json. Although this is deprecated, some indexes still include opm binaries that don't support podman default location for pull secret)

pre-cache/olm
```diff
@@ -35,7 +35,7 @@ export_index(){
     else
         opm_bin=$image_mount/bin/opm
     fi
-    $opm_bin index export -i $index -p $packages -c podman -f $index_download_path
+    $opm_bin render $index > $rendered_index_path
```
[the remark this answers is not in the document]
vitus133: Just to clarify, we don't include opm binary in our container image to keep its size minimal. The host doesn't contain this binary either. However, each OLM index image we pull contains opm binary capable of serving the registry and the index content, and we are copying this binary into the pre-caching workload container. So first of all, we can't bump it.
Regarding the supportability, all four default catalogs supported by Red Hat already contain opm binary supporting the `render` subcommand already in v4.9. This subcommand is not `alpha`.  Nevertheless you've put a seed of doubt in my mind. 
Maybe it's worth to keep both variants? First try with `render`, and if it fails try with `index export`?
What do you think?
I'll try to find since when `render` is supported to help us decide.

pre-cache/parse_index.py
```diff
@@ -27,54 +24,46 @@ def parse_args():
...
-    finally:
-        return images
+    for item in objects:
+        if item.get("schema") == "olm.bundle":
+            if item.get("name") in bundles and item.get("package") in packages:
+                images.extend([elem.get("image") for elem in item.get("relatedImages")])
+    return images
+
+def load_rendered_index():
+    with open("/tmp/index.json", "r") as f:
```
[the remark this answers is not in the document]
vitus133: fixed

pre-cache/parse_index.py
vitus133: Added exception handling

pre-cache/olm
```diff
@@ -62,37 +63,36 @@ olm_main(){
...
           log_debug "extract_pull_spec failed"
           return 1
         fi
         unmount_index $image_id
     done
-    if [[ $ret_val -ne 0 ]]; then
-        log_debug "Some of the OLM index extractions failed."
-    fi
-    return $ret_val
+    return 0
```
rcarrillocruz, line 95: not sure if this is needed, if loop finishes then the implicit return will be the last command of the function, so the return code should be implicitly zero I believe.

rcarrillocruz:
I don't have much context about the whole precache work, but this change looks ok to me, just a refactor to use a new parameter.

pre-cache/parse_index.py
```diff
@@ -27,59 +26,76 @@ def parse_args():
...
-        images = extract_images(bundles)
+        data = load_rendered_index()
+        images = extract_images(args, data)
         with open(args.img_list_file.name, args.img_list_file.mode) as f:
             f.write('\n'.join(images))
             f.write('\n')
-        exit(0)
+        sys.exit(0)
     except Exception as e:
         print(e)
```
danielmellado, line 99: I would use a logger here, rather than a print. Won't also be blocking the patch due to this but it's a good practice we may want to check for future reference.

danielmellado:
Same as Ricky, lacking context overall but the patch seems self-contained.

vitus133:
msg=Error: failed creating IAM Role (ci-op-xcqpz121-f9fd8-q5zqk-worker-role): LimitExceeded: Cannot exceed quota for RolesPerAccount: 1000
/override ci/prow/integration

PR #83, opened 2022-02-07
Fix related images yaml format

vitus133:
This commit fixes a cosmetic issue with default
yaml flow style used by python yaml
/cc @irinamihai @imiller0 @rcarrillocruz @danielmellado

### DIFF
--- a/hack/patch-precache-image.py
+++ b/hack/patch-precache-image.py
@@ -11,4 +11,4 @@ for item in obj.get('spec',{}).get('relatedImages',{}):
         item['image'] = pull
     break
 with open(file,'w') as f:
-    yaml.dump(obj, f)
+    yaml.dump(obj, f, default_flow_style=False)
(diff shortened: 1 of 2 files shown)


hack/patch-precache-image.py
```diff
@@ -11,4 +11,4 @@
         item['image'] = pull
     break
 with open(file,'w') as f:
-    yaml.dump(obj, f)
+    yaml.dump(obj, f, default_flow_style=False)
```
danielmellado, line 14: I assume you set default_flow_style to false so written yaml files would also have the block style.

vitus133: Right, so [config/manifests/pre-cache-image-patch.yaml](https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/83/files#diff-4c8b8ab4c6fed062d178c60b264310405d5163ec652a8535ca9ea2fde1affe12) is the written file. It is created when you run make bundle.
Thanks!

danielmellado: Awesome, thanks! ;)

PR #85, opened 2022-02-08
Bug 2051712: Extract spec from policies before precaching starts

vitus133:
This commit fixes a logic bug causing pre-caching
to get stuck without the software spec to pull.
/cc @irinamihai  @danielmellado @rcarrillocruz


controllers/precacheFsm.go
```diff
@@ -59,6 +59,20 @@ const (
...
 
+	specCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, "PrecacheSpecValid")
+	if specCondition == nil || specCondition.Status == metav1.ConditionFalse {
+		allManagedPoliciesExist, _, managedPoliciesPresent, err := r.doManagedPoliciesExist(ctx, clusterGroupUpgrade)
+		if err != nil {
+			return err
+		}
+
+		if allManagedPoliciesExist {
+			err = r.reconcileResources(ctx, clusterGroupUpgrade, managedPoliciesPresent)
```
irinamihai, line 70: Why do we also need this part?

rcarrillocruz: Yeah...I'm not sure I understand this PR, I must be missing context.
If precaching is for 'let's stage the artifacts for the upcoming upgrade', what is this reconcile policies call fixing?

irinamihai: Precaching needs information from the managed policies, but it also needs for the policies to be in the same namespace as the CGU. reconcileResources is creating copies of the managed policies in the namespace of the CGU, so pre-caching can access them. I discussed with Vitaly about the possibility of using doManagedPoliciesExist to get the namespaces of the managed policies instead of calling the reconcileResources.

vitus133: I think it's much better now.
Thanks folks

controllers/precacheFsm.go
```diff
@@ -60,6 +60,48 @@ func (r *ClusterGroupUpgradeReconciler) precachingFsm(ctx context.Context,
 	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) error {
 
 	r.setPrecachingRequired(clusterGroupUpgrade)
+	specCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, "PrecacheSpecValid")
```
irinamihai, line 63: Could we add this as a constant in the utils package?

controllers/precache.go
```diff
@@ -563,30 +520,33 @@ func (r *ClusterGroupUpgradeReconciler) getPrecacheJobTemplateData(
...
+	rv.Operators.PackagesAndChannels = spec.OperatorsPackagesAndChannels
+	return rv
+}
 
-	spec, err := r.getPrecachingSpec(ctx, clusterGroupUpgrade)
-	if err != nil {
-		return rv, err
-	}
-	r.Log.Info("[getPrecacheSoftwareSpec]", "PrecacheSoftwareSpec:", spec)
+// includeSoftwareSpecOverrides includes software spec overrides if present
```
irinamihai, line 538: Could we make this function comment a bit more explanatory? It's not clear from reading the description where the overrides are expected to be and where we're including them.

controllers/precache.go
```diff
@@ -334,24 +309,23 @@ func (r *ClusterGroupUpgradeReconciler) getPrecachingSpec(
 					return *new(ranv1alpha1.PrecachingSpec), nil
 				}
 				spec.PlatformImage = fmt.Sprintf("%s", image)
-				r.Log.Info("[getPrecachingSpec]", "ClusterVersion image", image)
+				r.Log.Info("[extractPrecachingSpecFromPolicies]", "ClusterVersion image", image)
 			case "Subscription":
```
irinamihai, line 313: It would be nice to have these cases as constants in utils: "ClusterVersion", "Subscription",...

samples/precache/catsrc-child-policy.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: policy.open-cluster-management.io/v1
```
irinamihai, line 1: For samples, would it be worth including the condensed subscription policy? I've also included it in one of the KUTTL testcases: https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/87/files#diff-413fcbbd4332dc448f6267751c3c9ebd0dd4214da59a4069b68f49eab514df8f.

Maybe the group-du-sno-validator-du-validator-policy also? Just to make sure we're covering a bunch of different policy formats.

controllers/precacheFsm.go
vitus133: Done

controllers/precache.go
vitus133: Done

controllers/precacheFsm.go
```diff
@@ -60,6 +60,48 @@ func (r *ClusterGroupUpgradeReconciler) precachingFsm(ctx context.Context,
...
+		allManagedPoliciesExist, _, managedPoliciesPresent, err := r.doManagedPoliciesExist(ctx, clusterGroupUpgrade)
+		if err != nil {
+			return err
+		}
+		if !allManagedPoliciesExist {
+			meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
+				Type:    "PrecacheSpecValid",
+				Status:  metav1.ConditionFalse,
+				Reason:  "NotAllManagedPoliciesExist",
+				Message: "Not all managed policies specified in the CGU are present on the hub cluster"})
```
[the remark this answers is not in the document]
vitus133: Good point, done

controllers/precache.go
vitus133: Done

samples/precache/catsrc-child-policy.yaml
vitus133: There is a subscription policy (Line 40):
https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/85/files#diff-35155d9baeb86ab3c1c93fd8a3f226eae04c83605802a4673ae3566c944ea273
I added only policies relevant for pre-caching, since these are pre-caching specific tests. Maybe we should think about adding e2e tests that will test pre-caching and upgrade together? If / when we do, I agree it would make much more sense adding all the policy kinds you've mentioned.

PR #86, opened 2022-02-08
Fix policy content parsing

irinamihai:
Description:
- handle the case where the policy template content is missing  metadata fields: if either kind, metadata.name or   metadata.namespace is missing, log the info and move to the next content template
- filter only the Subscription type templates

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -909,15 +909,16 @@ func (r *ClusterGroupUpgradeReconciler) getPolicyContent(
 			objectDefinitionMetadataContent := innerObjectDefinitionContent["metadata"].(map[string]interface{})
 			// Save the kind, name and namespace if they exist and if kind is of Subscription type.
 			// If kind is missing, log and skip.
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -898,19 +898,61 @@ func (r *ClusterGroupUpgradeReconciler) getPolicyContent(
...
+			// If kind is missing, log and skip.
+			_, ok := innerObjectDefinitionContent["kind"]
+			if ok == false {
+				r.Log.Info(
+					"[getPolicyContent] Policy is missing its spec.policy-templates.objectDefinition.spec.object-templates.kind",
+					"policyName", managedPolicyName)
+				continue
+			}
+			// Filter only Subscription templates.
+			if innerObjectDefinitionContent["kind"] != utils.PolicyTypeSubscription {
```
rcarrillocruz, line 920: you lookup innerObjectDefinitionContent["kind"] twice, in line 912 and 920. 
You could maybe store it in a 'kind' variable in 912 so it can be used in 920.

PR #87, opened 2022-02-11
Add KUTTL test for operator upgrade

irinamihai:


deploy/acm/policies/all_policies/policy5-subscriptions.yaml
```diff
@@ -0,0 +1,208 @@
...
+                        apiVersion: operators.coreos.com/v1alpha1
+                        kind: Subscription
+                        metadata:
+                            annotations:
+                                ran.openshift.io/ztp-deploy-wave: "2"
+                            name: local-storage-operator
+                            namespace: openshift-local-storage
+                        spec:
+                            channel: "4.9"
+                            installPlanApproval: Automatic
```
Missxiaoguo, line 204: I think installPlan will only be approved when its Approval is "Manual"? I know it doesn't really matter of the value of installPlanApproval set here for testing, as you are patching the installPlans with Approval Manual in the created MCVs later,  but I feel a bit confused when seeing the sub has installPlanApproval with Automatic but its installplan finally cat get approved. I think better to provide the example sub yaml that match the logic? Maybe explicitly add the installPlanApproval: Manual in other subs as well because the default is Automatic.

tests/kuttl/tests/operator-upgrade/00-assert.yaml
```diff
@@ -0,0 +1,192 @@
...
+kind: ManagedClusterView
+metadata:
+  name: cgu-default-subscription-ptp-operator-subscription
+  namespace: spoke5
+spec:
+  scope:
+    name: ptp-operator-subscription
+    namespace: openshift-ptp
+    resource: subscriptions.operators.coreos.com
+---
```
Missxiaoguo, line 181: Looks like sriov-network-operator-subscription MCVs for spoke2 and spoke5 are missing

PR #89, opened 2022-02-17
Get image payload for a given a version from the update graph

rcarrillocruz:
Up till now the expectation for precaching was that the ClusterVersion
would pass an image, but it should also accept a version.


controllers/precache.go
```diff
@@ -298,18 +346,32 @@ func (r *ClusterGroupUpgradeReconciler) extractPrecachingSpecFromPolicies(
...
+				} else {
+					upstream := object["spec"].(map[string]interface {
+					})["upstream"].(string)
+					channel := object["spec"].(map[string]interface {
+					})["channel"].(string)
+					version := object["spec"].(map[string]interface {
+					})["desiredUpdate"].(map[string]interface{})["version"].(string)
+					image, err := r.getImageForVersionFromUpdateGraph(upstream, channel, version)
+					if err != nil {
+						return spec, err
```
vitus133, line 370: This might fail for variety of reasons:
- User left `upstream`, `channel` or `version` empty. 
- User made a typo in one of them
- The upstream server failed, or there is a temporary network issue
I wonder, would it be better to raise a condition (utils.PrecacheSpecValidCondition) and pass the error message as message?
This will give user a chance to fix the policy without digging in the log

vitus133 (commented):
Hey,
Just one comment about a possible error handling improvement.
Otherwise - awesome!
Thanks

controllers/precache.go
rcarrillocruz: Good question, I do not know what is coming from ZTP.
Are you saying that it may not be present those fields?
What is the contract?

rcarrillocruz: If the answer is 'we can't assume anything' that's fine, i can check and if not present then return nil spec as it's been till now

controllers/precache.go
```diff
@@ -298,18 +346,43 @@ func (r *ClusterGroupUpgradeReconciler) extractPrecachingSpecFromPolicies(
...
+						msg := "Platform image could not be inferred, either upstream/channel/version are missing or upstream endpoint is unreachable"
+						meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
+							Type:    utils.PrecacheSpecValidCondition,
+							Status:  metav1.ConditionFalse,
+							Reason:  "PlatformImageInvalid",
+							Message: msg})
+						return *new(ranv1alpha1.PrecachingSpec), nil
+					}
+
+					spec.PlatformImage = image
```
vitus133, line 383: Could it be better if we just pass the actual error as the condition message?
```suggestion
					image, err := r.getImageForVersionFromUpdateGraph(upstream, channel, version)
					if err != nil {
						meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
							Type:    utils.PrecacheSpecValidCondition,
							Status:  metav1.ConditionFalse,
							Reason:  "PlatformImageInvalid",
							Message: err.Error()})
						return *new(ranv1alpha1.PrecachingSpec), nil
					}
					spec.PlatformImage = image
```

PR #96, opened 2022-02-25
Add cleanup finalizer for ClusterGroupUpgrade CRs

irinamihai:
Description:
- Bug 2054421
- add cleanup finalizer for ClusterGroupUpgrade CRs so that managedClusterViews created in different namespaces than the CGU are deleted when the CGU is deleted.
- upon reconciling, check if CGU is marked for deletion and if so, look at the cleanup finalizer. If the cleanup finalizer is still
  present, do the needed cleanup and remove the finalizer.
- if the cleanup finalizer doesn't exist, add it


vitus133 (approved):
Looks good!
/lgtm

PR #97, opened 2022-02-25
Bug 2058414: Support ClusterVersion without "desiredUpdate"

vitus133:
Support pre-caching ignoring clusterversion policies where
the desiredUpdate field is omitted.

/cc @irinamihai @imiller0 @nishant-parekh  @Missxiaoguo

### DIFF
--- a/controllers/precache.go
+++ b/controllers/precache.go
@@ -344,9 +344,13 @@ func (r *ClusterGroupUpgradeReconciler) extractPrecachingSpecFromPolicies(
 			kind := object["kind"]
 			switch kind {
 			case utils.PolicyTypeClusterVersion:
-				image := object["spec"].(map[string]interface {
-				})["desiredUpdate"].(map[string]interface{})["image"]
-				if image != nil {
+				cvSpec := object["spec"].(map[string]interface{})
+				desiredUpdate, found := cvSpec["desiredUpdate"]
+				if !found {
+					continue
+				}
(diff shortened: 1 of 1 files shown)


controllers/precache.go
```diff
@@ -344,9 +344,13 @@ func (r *ClusterGroupUpgradeReconciler) extractPrecachingSpecFromPolicies(
...
-				image := object["spec"].(map[string]interface {
-				})["desiredUpdate"].(map[string]interface{})["image"]
-				if image != nil {
+				cvSpec := object["spec"].(map[string]interface{})
+				desiredUpdate, found := cvSpec["desiredUpdate"]
+				if !found {
+					continue
+				}
+				image, found := desiredUpdate.(map[string]interface{})["image"]
+				if found && image != "" {
```
yliu127, line 353: We need to do value check on these parameters as well. 
				} else {
					upstream := object["spec"].(map[string]interface {
					})["upstream"].(string)
					channel := object["spec"].(map[string]interface {
					})["channel"].(string)
					version := object["spec"].(map[string]interface {
					})["desiredUpdate"].(map[string]interface{})["version"].(string)

vitus133: Yes, they are checked inside function `getImageForVersionFromUpdateGraph`. If one of them is omitted, PrecacheSpecValidCondition is set to false with reason PlatformImageInvalid

PR #100, opened 2022-03-03
BZ #2055309 - Include PKI folder when precaching against untrusted registries

alosadagrande:
This PR tries to fix the problem I found when running precache task against an untrusted registry. My environment is disconnected so the registry and so the upgrade graph.

```
oc adm upgrade
Upstream: http://graph.apps.cnf20.cloud.lab.eng.bos.redhat.com/api/upgrades_info/v1/graph
```

When creating a CGU that enables precaching the precache job does not have the K8S node certificates available. I believe that the precache task contacts the untrusted registry before chrooting.

```
$ oc logs -f pre-cache--1-plmqp
Error: Error initializing source docker://fx2-1a.cloud.lab.eng.bos.redhat.com:8443/test/release@sha256:fd96300600f9585e5847f5855ca14e2b3cafbce12aefe3b3f52c5da10c4476eb: error pinging docker registry fx2-1a.cloud.lab.eng.bos.redhat.com:8443: Get "https://fx2-1a.cloud.lab.eng.bos.redhat.com:8443/v2/": x509: certificate signed by unknown authority`
```

Basically, I just mount the directory where the K8S node has the untrusted registry certificates into the precache container image using a hostPath since a hostPath mounting /host is already configured. Once done, the precache works as expected

Signed-off-by: Alberto Losada <alosadag@redhat.com>

### DIFF
--- a/pre-cache/copy-env.sh
+++ b/pre-cache/copy-env.sh
@@ -19,4 +19,5 @@ done
 # Copy containers policy and cacert of disconnected registries if present
 cp /host/etc/containers/policy.json /etc/containers/policy.json
 cp -a /host/etc/docker /etc/ || true
-
+cp -a /host/etc/pki/ca-trust/source/anchors /etc/pki/ca-trust/source/ && update-ca-trust || true
+cp -a /host/etc/pki/ca-trust/extracted/pem /etc/pki/ca-trust/extracted/ || true
--- a/pre-cache/olm
+++ b/pre-cache/olm
@@ -24,7 +24,7 @@ export_index(){
     # copy pull secret to the default runtime dir - opm doesn't accept authfile as an arg
     [ ! -d "$podman_pull_path" ] && mkdir -p "$podman_pull_path"
     cp $pull_secret_path "$podman_pull_path/auth.json"
-    capath="/etc/docker/certs.d/${index%/*}/ca.crt"
+    capath="/host/etc/docker/certs.d/${index%/*}/ca.crt"
     # opm doesn't accept cacert as an arg
     if [[ -f "$capath" ]]; then
         cp $capath /etc/pki/ca-trust/source/anchors/ 


vitus133:
/hold
I wish it was so simple, but it's the same issue that caused me to more or less rewrite the whole thing.
You can't mount a host path to the same path in the container, as is contradicts the OCP documentation:
https://docs.openshift.com/container-platform/4.9/storage/persistent_storage/persistent-storage-hostpath.html

> Do not mount to the container root, /, or any path that is the same in the host and the container. This can corrupt your host system if the container is sufficiently privileged. It is safe to mount the host by using /host. The following example shows the / directory from the host being mounted into the container at /host.

It's fine for unblocking your testing, but not for production... Unfortunately, as it could have make things much simpler.

alosadagrande:
How can be corrupted if the certificates folder is mounted as readOnly?

vitus133:
> How can be corrupted if the certificates folder is mounted as readOnly?

I don't know, probably because of a bug or a limitation in a storage driver? And `readOnly` or not - I think it shouldn't matter, because the warning text quoted above doesn't leave any freedom of interpretation. I can't estimate the risk, so I wouldn't do in production something that OCP docs explicitly tell me not to do.

alosadagrande:
What about copying the /host/etc/pki/ca-trust/extracted/pem/ files into /etc/pki/ca-trust/extracted/pem/ folder?

Similar to this: https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/pre-cache/copy-env.sh#L21

vitus133:
> What about copying the /host/etc/pki/ca-trust/extracted/pem/ files into /etc/pki/ca-trust/extracted/pem/ folder?
> 
> Similar to this: https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/pre-cache/copy-env.sh#L21

I think this should do the trick. Could you please coordinate this  with @irinamihai? She is already trying something in this direction. If you can test on your env, it could be great

PR #102, opened 2022-03-07
Allow precaching to obtain compliant managed policies

irinamihai:
Description:
- Split doManagedPoliciesExist into a function that is only looking at policies and a separate one that is creating managedClusterViews for Subscriptions
- Add parameter to doManagedPoliciesExist to allow (not) filtering the NonCompliant policies
- Update precaching to call doManagedPoliciesExist with the new parameter for not filtering NonCompliant policies thus allowing
  Compliant policies to be parsed in order to obtain catalog source info


vitus133:
Hey, looks good!
/lgtm

PR #104, opened 2022-03-08
Update vendor to fix violations

danielmellado:
This commit fixes go vet vendor violations, such as in explicitly
required in go.mod, but not marked as explicit in vendor/modules.


rcarrillocruz:
This looks bogus, we don't have vendored dependencies.

PR #105, opened 2022-03-08
Rename UpgradeCannotStart for UpgradeBlockedOnCR

rcarrillocruz:
The UpgradeCannotStart is too similar to UpgradeNotStarted, which
could confuse users.
Since the semantics of this state is to signal that an upgrade
cannot start due to being blocked by another CR which needs to complete
before moving on, just rename it to UpgradeBlockedOnCR.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -135,7 +135,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		if readyCondition.Reason == "PrecachingRequired" {
 			requeueAfter := 5 * time.Minute
 			nextReconcile = ctrl.Result{RequeueAfter: requeueAfter}
-		} else if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == "UpgradeCannotStart" {
+		} else if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == "UpgradeBlockedOnCR" {
 			// Before starting the upgrade check that all the managed policies exist.
 			allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent, err :=
 				r.doManagedPoliciesExist(ctx, clusterGroupUpgrade, true)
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -135,7 +135,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		if readyCondition.Reason == "PrecachingRequired" {
 			requeueAfter := 5 * time.Minute
 			nextReconcile = ctrl.Result{RequeueAfter: requeueAfter}
-		} else if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == "UpgradeCannotStart" {
+		} else if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == "UpgradeBlockedOnCR" {
```
Missxiaoguo, line 138: I think we should create a constants variable for it.

controllers/clustergroupupgrade_controller.go
```diff
@@ -222,7 +222,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
 					Type:    "Ready",
 					Status:  metav1.ConditionFalse,
-					Reason:  "UpgradeCannotStart",
+					Reason:  "UpgradeBlockedOnCR",
```
Missxiaoguo, line 225: UpgradeBlockedOnCR is not accurate for this case - cannot start due to missing policies.

Probably something like "UpgradeNotReadyToStart"?

rcarrillocruz: Huh, yes, I did a find/replace.
So this means this particular UpgradeCannotStart has had a dual meaning all this time.
I need to think what's the best transition here.

controllers/clustergroupupgrade_controller.go
rcarrillocruz: Sure, I can do on a separate PR.

irinamihai:
@rcarrillocruz, KUTTL tests need to be updated. Sorry...

PR #106, opened 2022-03-08
Fix gosimple violations

danielmellado:
This commit fix all occurrences of S1002 violation, avoiding comparisons
to bool constants.


nishant-parekh (approved):
I think we should have a linter job to check that we do not introduce these in the future.. something to consider..

PR #113, opened 2022-03-11
Add golangci-lint

danielmellado:
This commit adds golangci-lint and fixes multiple violations. Some of
them have just been skipped by using nolint directives.

It also adds a new hack/update-deps script to update vendor
dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,11 @@ fmt: ## Run go fmt against code.
 	@echo "Running go fmt"
 	go fmt ./...
 
+.PHONY: golangci-lint
+golangci-lint: ## Run golangci-lint against code.
+	@echo "Running golangci-lint"
+	hack/golangci-lint.sh
+
 .PHONY: vet
 vet: ## Run go vet against code.
 	@echo "Running go vet"
@@ -136,7 +141,7 @@ common-deps-update:	controller-gen kustomize
 	go mod tidy
 
 .PHONY: ci-job
-ci-job: common-deps-update fmt vet lint unittests
+ci-job: common-deps-update fmt vet lint golangci-lint unittests
 
 .PHONY: shellcheck
 shellcheck: ## Run shellcheck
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -101,6 +101,11 @@ fmt: ## Run go fmt against code.
 	@echo "Running go fmt"
 	go fmt ./...
 
+.PHONY: golangci-lint
+golangci-lint: ## Run golangci-int against code.
```
donpenney, line 105: nit: typo in comment

Makefile
```diff
@@ -101,6 +101,11 @@ fmt: ## Run go fmt against code.
...
+	@echo "Running golangci-lint"
+	HOME=/tmp hack/golangci-lint.sh
```
donpenney, line 107: There's already a hack/lint.sh and Make target. Any reason to not just add golangci-lint to that existing script? That script could just become the "general purpose" linter tool, possibly growing in the future to include other linters like pylint, shellcheck, bashate, etc.

Makefile
danielmellado: Thanks!

Makefile
danielmellado: Well, basically I put that out there in order to have a small script per linter. i.e. golint allows you to have the `-set-exit-status param`which is used in that bash script but it wouldn't apply to golangci-lint nor to anyone else. What we could do, though, is to aggregate such scrips calls into one while keeping them separate into files.

controllers/precacheFsm.go
```diff
@@ -200,11 +201,14 @@ func (r *ClusterGroupUpgradeReconciler) handleNotStarted(ctx context.Context,
 
 // handlePreparing handles conditions in PrecacheStatePreparingToStart
 // returns: error
+//nolint:unparam
 func (r *ClusterGroupUpgradeReconciler) handlePreparing(ctx context.Context,
 	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade,
 	cluster string) (string, error) {
 
+	//nolint:ineffassign
 	currentState, nextState := PrecacheStatePreparingToStart, PrecacheStatePreparingToStart
```
vitus133, line 210: I wonder why do I have ineffectual assign here

vitus133, line 210: I wonder why do I have ineffectual assign here

PR #115, opened 2022-03-15
Pre-caching: Add checks for return values

irinamihai:
Description:
- add return value checks for pre-caching bash scripts so that the pre-caching job doesn't report false completion
- BZ#2063933


vitus133 (approved):
Thanks, looks good!

PR #121, opened 2022-03-16
Fix typo in README

danielmellado:
This commit fixes a typo on the readme that points out to
deleteObjectsOnCompletion where it should be delteObjects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ following states:
   * In this state, the controller will periodically check if all the policies for the **ClusterGroupUpgrade** are compliant, and in that case it will transition to **UpgradeCompleted**. This is to give a chance for upgrades to catch up, as they could be taking long to complete due to network, CPU or other issues but they are not really stuck and they can indeed complete
 * **UpgradeCompleted**
   * In this state, the upgrades of the clusters are complete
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ following states:
   * In this state, the controller will periodically check if all the policies for the **ClusterGroupUpgrade** are compliant, and in that case it will transition to **UpgradeCompleted**. This is to give a chance for upgrades to catch up, as they could be taking long to complete due to network, CPU or other issues but they are not really stuck and they can indeed complete
 * **UpgradeCompleted**
   * In this state, the upgrades of the clusters are complete
-  * If the *deleteObjectsOnCompletion* field is set to **true**, the controller will delete the underlying RHACM policies. This is to avoid having RHACM Hub to continously check for compliance since the upgrade has been successful
+  * If the *deleteObjects* field is set to **true**, the controller will delete the underlying RHACM policies. This is to avoid having RHACM Hub to continously check for compliance since the upgrade has been successful
```
Missxiaoguo, line 45:
```suggestion
  * If the *actions.afterCompletion.deleteObjects* field is set to **true** which is the default value, the controller will delete the underlying RHACM policies. This is to avoid having RHACM Hub to continously check for compliance since the upgrade has been successful
```

PR #122, opened 2022-03-16
Add go-bindata module dependency

donpenney:
The go-bindata package supports the generation of Go source code from
a static file. This mechanism is used by the upgrade recovery image to
build the recovery utility into the image for delivery to the spoke as
part of taking the pre-upgrade backup.


danielmellado:
I assume you'll be landing a follow-up which will be using this dependency

donpenney:
> 

Yes, I'm breaking down the integration work into multiple smaller PRs for easier review.

PR #123, opened 2022-03-17
Set default backup value to false

danielmellado:
This commit sets the kubebuilder default value to false in the api.


donpenney (approved):
/lgtm

PR #124, opened 2022-03-17
new templates folder for pre-cache and backup job templates

sabbir-47:
For [CNF-4055](https://issues.redhat.com/browse/CNF-4055) story, it aims to refactor and reorganize codes around pre-caching to be reused by backup implementation.
This pull requests of sub-task 4257 reorganizes templates to a template folder and adds backup templates to be used by backup trigger implementation.

### DIFF
--- a/controllers/managedClusterResources.go
+++ b/controllers/managedClusterResources.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"text/template"
 
+	"github.com/openshift-kni/cluster-group-upgrades-operator/controllers/templates"
+
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


vitus133:
I tested the pre-caching part and it works as expected. Could it be possible to squash the commits into one? This will make it easier to cherrypick if we decide to.

danielmellado:
Yeah, squashing that into one commit would be somehow a good practice and I highly recommend it. Not only for reviewing but also as Vitaly suggest for cherry-picking ;)

danielmellado:
Also this time it's only 2 commit, so please fixup the most recent one fixing the violations into previous one, thanks!

sabbir-47:
Squashed into single commit as suggested @vitus133  @danielmellado

PR #125, opened 2022-03-17
Introduce upgrade recovery tool

donpenney:
This update introduces the upgrade recovery tool, which is written to
the recovery partition as part of taking the pre-upgrade backup. This
utility has functions for backing up data, as well as for managing the
recovery from failure. The go-bindata module is used to generate go
source code encapsulating the script file. The source code is
generated or updated by running the "make update-bindata" command.

In addition, this update adds support for running shellcheck, a linter
for bash scripts.


danielmellado:
let's wait on CI

donpenney:
Makefile has been updated to resolve issue hit by the integration CI test, which is now passing.

PR #126, opened 2022-03-17
Introduce upgrade recovery image

donpenney:
This update introduces the upgrade recovery image, which is
responsible for writing the upgrade recovery script to the recovery
partition and taking the pre-upgrade backup.

Co-Authored-By: Daniel Mellado <dmellado@redhat.com>
Co-Authored-By: Yolanda Robla <yroblamo@redhat.com>
Co-Authored-By: Don Penney <dpenney@redhat.com>


donpenney:
/hold
This PR depends on merge of https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/125

donpenney:
Note that the current test failures are due to the missing dependency being introduced by #125 ... Once that PR merges, these test should be able to pass.

recovery/cmd/launchBackup.go
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright 2021 Red Hat, Inc.
```
irinamihai, line 2: 2022?

donpenney: Code was started last year in a dev repo. I meant to check for dates... I'll fix these up now

recovery/cmd/root.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.openshift-ai-image-backup.yaml)")
```
sabbir-47, line 59: I'm not sure if we need this persistent flag.

recovery/cmd/root.go
```diff
@@ -0,0 +1,87 @@
...
+
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
```
sabbir-47, line 64: Also, i think we are not reading anything from config file or passing any environment variable, so i'm not sure if we need it

sabbir-47 (commented):
Please have a look at 2 comments that i've added

recovery/cmd/root.go
donpenney: I wondered about these when I was integrating the code, but didn't want to modify too much. Once #125 merges, though, I will update this PR with this code dropped.

recovery/cmd/root.go
donpenney: Removed

recovery/cmd/root.go
donpenney: Removed

sabbir-47 (commented):
/lgtm

PR #129, opened 2022-03-21
Bump Go to 1.17

rcarrillocruz:


donpenney:
It would probably be good for us to add a validation check in the Makefile to ensure developers have the required go version installed/running. Otherwise, running "go mod vendor/tidy" commands could lead to someone inadvertently pushing changes that conflict with 1.17

danielmellado:
As of now, CI complains about `go mod tidy: go.mod file indicates go 1.17, but maximum supported version is 1.16`. After checking this in https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/cluster-group-upgrades-operator/openshift-kni-cluster-group-upgrades-operator-main.yaml#L10 it points out that it's using go 1.16 for main branch. We'll probably have to bump that before merging this change.

danielmellado:
Just posted a commit to do so at https://github.com/openshift/release/pull/27143

danielmellado:
merged, let's unhold this

PR #131, opened 2022-03-22
[cnf-4257] precaching implementation reorganization for reusability

sabbir-47:
This pull request aims at minimal refactoring of precaching implementation to be able to be reused by triggering backup in coming days. Few functions which are related to managedCluster have been moved from precache.go to managedClusterResources.go which are more or less a bit generic functions. Some functions signature have been changed as well.


danielmellado:
As with last PR comment, would you mind squashing this up? Thanks!

danielmellado:
waiting on #124 to get merged

vitus133:
/unhold
(as #124 has merged)

controllers/managedClusterResources.go
```diff
@@ -250,3 +260,306 @@ func (r *ClusterGroupUpgradeReconciler) renderYamlTemplate(
...
+//          using a set of templated manifests
+// returns: error
+func (r *ClusterGroupUpgradeReconciler) deployWorkload(
+	ctx context.Context,
+	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade,
+	cluster, workloadType, mcvTemplate string, template []resourceTemplate) error {
+
+	// the idea is to reuse the deployworkload function for backup job, so using switch can different condition different jobs
+	var spec *templateData
+	switch workloadType {
```
donpenney, line 526: golangci-lint test is failing here, suggesting an if-check rather than a single-case switch

donpenney (changes requested):

controllers/managedClusterResources.go
sabbir-47: i will add default for now, the idea was, with the next PR there will be another workloadtype "backup"

rcarrillocruz:
I would not put the precache functions into managedClusterResources.go. Unless they are really generic,
it makes more sense to me to leave precaching specific functions to a file that points that is code related to precaching.
e.g. 

getPrecacheJobTemplateData
precachingCleanup

Also, the other functions despite their name not specifying they are specific to precache functionality, they seem per the comments they are for precaching.

So, if there is code that can be reused by backup and precache I would suggest put it under a utils.go, managedClusterResources.go doesn't seem very descriptive to what their functions do.

Regarding the change for removal of 'Precache' prefix on conditions, is your intent to reuse those for backup?

sabbir-47:
> precachingCleanup

This is a generic function, backup can use this function and the name will change to cleanup.

> getPrecacheJobTemplateData

This is also a generic function, backup can use this function to set templates data and the name will change to getJobTemplateData

>Regarding the change for removal of 'Precache' prefix on conditions, is your intent to reuse those for backup?

yes

>So, if there is code that can be reused by backup and precache I would suggest put it under a utils.go, managedClusterResources.go doesn't seem very descriptive to what their functions do.

If we are really looking at refactoring in a logical way, i would like to point out on precaching constants both reside at utils/constants.go and precacheFsm.go. what is the utils/common.go file is about? Why are we keeping overrides.go file under controllers having a single function? Do we also look to address them?

rcarrillocruz:
Ok, if those functions will be used as is for backup then it's good to move to a common file and change their name to more generic as you are doing.

> If we are really looking at refactoring in a logical way, i would like to point out on precaching constants both reside at utils/constants.go and precacheFsm.go. what is the utils/common.go file is about? Why are we keeping overrides.go file under controllers having a single function? Do we also look to address them?

That is bad IMHO, we should just have one file to have constants related to a particular Go package and a utils/common that is common to all.
I think we could split controllers folder into multiple packages.
Like:

controllers\
  clustergroupupgrade\
  managedcluster\
  precache\
  precache\templates\
  backup\
  backup\templates\
  utils\

sabbir-47:
Should be a focus for future refactor of the codes, this is more scoped to reorganize precache implementation as we discussed

controllers/managedClusterResources.go
sabbir-47: fixed and squashed to previous commit.

PR #133, opened 2022-03-29
Bug 2069569: pre-caching: fix opm authorization

vitus133:
This fixes an authorization problem occurring when
trying to render an OLM index container with opm.
There are two tools used in the precaching workload
from inside the container, podman and opm.
Podman is using auth.json from /run/containers/0
when run as root. Opm is using ~/.docker/config.json.
This change adds missing pull secret copy to the
latter for the opm use, and also puts all the relevant
copy operations to the "copy_env.sh" script
/cc @irinamihai @rcarrillocruz @danielmellado 


vitus133:
msg=Error: creating IAM instance profile ci-op-3g4k81si-e1ac0-kzr6g-worker-profile: LimitExceeded: Cannot exceed quota for InstanceProfilesPerAccount: 1000
/override ci/prow/e2e-aws

vitus133:
msg=Error: creating IAM instance profile ci-op-3g4k81si-f9fd8-wtg8j-worker-profile: LimitExceeded: Cannot exceed quota for InstanceProfilesPerAccount: 1000
/override ci/prow/integration

pre-cache/copy-env.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+cwd="${cwd:-/opt/precache}"
```
rcarrillocruz, line 3: what is cwd here? shouldn't you use pwd?

pre-cache/common
```diff
@@ -4,6 +4,10 @@ container_tool="${container_tool:-podman}"
 pull_secret_path="${pull_secret_path:-/host/var/lib/kubelet/config.json}"
 pull_spec_file="${pull_spec_file:-/tmp/images.txt}"
 config_volume_path="${config_volume_path:-/etc/config}"
+podman_pull_path="${podman_pull_path:-/run/containers/0}"
```
rcarrillocruz, line 7: i see this is a refactor so this should be working, but I'm curious what does /run/containers/0 contain? Is this a file that always exist on coreos in the underlying FS of an openshift cluster?

pre-cache/copy-env.sh
vitus133: No, it is for unit testing:
https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/04a3b444fc20580d28d41cd266452e0252d2435c/Makefile#L348

pre-cache/common
danielmellado: when you do a podman login, the auth is stored there, so I assume it's just creating the path.

vitus133: No, this is a folder where podman that runs as root looks for the pull secret by default. We copy the pull secret there from the host

pre-cache/olm
```diff
@@ -17,13 +13,9 @@ extract_pull_spec(){
 
 render_index(){
     index=$1
-    pull_secret_path=$2
```
danielmellado, line 20: I understand that pull_secret_path is no longer needed because it would use the default or something. Overall, would you please put some description on the PR/commit message? I think it'd be good for the future us. Thanks!

vitus133: Done

vitus133:
msg=Error: error creating EC2 VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.
/override ci/prow/e2e-aws

vitus133:
msg=Error: Error creating VPC Endpoint: VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached
/override ci/prow/integration

PR #136, opened 2022-03-29
Fix OLM index extraction

vitus133:
This fixes a bug in sqlite OLM catalog extraction, causing
operator bundle to be identified incorrectly. The bug was
percieved as a failure to pull one or more images.
The problem is fixed by cherry-picking this commit from the
main branch:

Commit d019df2490ac8b0753b953abd8a15a09e86f6ed8
    This fixes an authorization problem occurring when
    trying to render an OLM index container with opm.
    There are two tools used in the precaching workload
    from inside the container, podman and opm.
    Podman is using auth.json from /run/containers/0
    when run as root. Opm is using ~/.docker/config.json.
    This change adds missing pull secret copy to the
    latter for the opm use, and also puts all the relevant
    copy operations to the "copy_env.sh" script

/cc @irinamihai @rcarrillocruz @danielmellado

### DIFF
--- a/pre-cache/copy-env.sh
+++ b/pre-cache/copy-env.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+cwd="${cwd:-/opt/precache}"
+. $cwd/common
+
 # Copy podman dependencies to the container. Podman itself runs off the host in the
 # container by adding /host/usr/bin to the path
 # This is a work around for podman not passing LD_LIBRARY_PATH environment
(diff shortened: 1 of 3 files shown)


vitus133:
This should also fix https://bugzilla.redhat.com/show_bug.cgi?id=2066441

vitus133:
msg=Error: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached
/override ci/prow/e2e-aws 
/override ci/prow/integration

pre-cache/copy-env.sh
```diff
@@ -22,3 +25,8 @@ cp /host/etc/containers/registries.conf /etc/containers/registries.conf || true
 cp -a /host/etc/docker /etc/ || true
 cp -a /host/etc/pki/ca-trust/source/anchors /etc/pki/ca-trust/source/ && update-ca-trust || true
 cp -a /host/etc/pki/ca-trust/extracted/pem /etc/pki/ca-trust/extracted/ || true
+[ ! -d "$podman_pull_path" ] && mkdir -p "$podman_pull_path"
+cp $pull_secret_path "$podman_pull_path/auth.json"
+[ ! -d "$docker_pull_path" ] && mkdir -p "$docker_pull_path"
+cp $pull_secret_path "$podman_pull_path/config.json"
```
donpenney, line 31: Is this meant to be $docker_pull_path? It looks odd to have it after the docker_pull_path dir check otherwise

donpenney, line 31: Is this meant to be $docker_pull_path? It looks odd to have it after the docker_pull_path dir check otherwise

vitus133: Thanks, this is a good catch!

PR #137, opened 2022-03-29
Properly update policy/placement binding/rule objects

jc-rh:
Bug 2069681

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -850,6 +850,9 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyNameForCopiedPo
 func (r *ClusterGroupUpgradeReconciler) createNewPolicyFromStructure(
 	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, policy *unstructured.Unstructured) error {
 
+	if err := controllerutil.SetControllerReference(clusterGroupUpgrade, policy, r.Scheme); err != nil {
+		return err
+	}
 	existingPolicy := &unstructured.Unstructured{}
 	existingPolicy.SetGroupVersionKind(schema.GroupVersionKind{
 		Group:   "policy.open-cluster-management.io",
@@ -863,9 +866,7 @@ func (r *ClusterGroupUpgradeReconciler) createNewPolicyFromStructure(
 
 	if err != nil {
 		if errors.IsNotFound(err) {
-			if err := controllerutil.SetControllerReference(clusterGroupUpgrade, policy, r.Scheme); err != nil {
-				return err
-			}
+
 			err = r.Client.Create(ctx, policy)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -849,16 +849,24 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyNameForCopiedPo
 
 func (r *ClusterGroupUpgradeReconciler) createNewPolicyFromStructure(
 	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, policy *unstructured.Unstructured) error {
+
+	if err := controllerutil.SetControllerReference(clusterGroupUpgrade, policy, r.Scheme); err != nil {
```
irinamihai, line 853: Do we really need to SetControllerReference each time we are calling this function? To me, it seems that doing this upon creation is enough.

controllers/clustergroupupgrade_controller.go
```diff
@@ -984,28 +993,26 @@ func (r *ClusterGroupUpgradeReconciler) getPolicyContent(
 
 func (r *ClusterGroupUpgradeReconciler) ensureBatchPlacementRule(ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, prName string) error {
 
+	pr := r.newBatchPlacementRule(clusterGroupUpgrade, prName)
+
+	if err := controllerutil.SetControllerReference(clusterGroupUpgrade, pr, r.Scheme); err != nil {
```
irinamihai, line 998: Same comment as above.

controllers/clustergroupupgrade_controller.go
jc-rh: This is actually the fix for the kuttl test failures, where updated objects get left behind causing subsequent testcases (with same CGU name) to fail. Before we use retrieved object to do update, that's why this is only needed in create case. Now that we do the real update with the desired policy object we just create, we have to set a few things like owner, resource version for the update to work.

serngawy (commented):
I'm wondering how TALO will behave if the policy get updated after it was complaint and TALO started the remediation of the next policy ? Does TALO will start over again ? Is that will make any issue ?

jc-rh:
> I'm wondering how TALO will behave if the policy get updated after it was complaint and TALO started the remediation of the next policy ? Does TALO will start over again ? Is that will make any issue ?

Good point. As you probably already know, we only update policies when the CGU state is "upgradeNotStarted" or before. As soon as the state moves to "upgradeNotCompleted", we don't ever look back and update them again. So the upgrade would be stuck somewhere if a change is made to managed policies, whether the affected clusters already remediated or not. @irinamihai , is this well documented in customer doc? I suppose the recovery is as easy as delete and recreate CGU.

A potential improvement that we can do in near future is to monitor changes on managed policies once upgrade is started. If a change detected, we can abort with a clear error message in CGU status.

PR #139, opened 2022-03-30
Delete cgu-controller-manager serviceAccount yaml.

danielmellado:
Newer operator-sdk versions explicitly copied the SA into the bundle
directory. That behavior has changed and now it's getting created
directly naming it from in the CSV.

As the operator-sdk behavior changed, now its validation fails unless we
explicitly delete this formerly valid  SA.


donpenney (commented):
/lgtm

PR #140, opened 2022-03-31
Fix docker pull secret path

vitus133:
This fixes a typo in the docker pull path.
Until fixed, OLM index processing can fail
in disconnected environments


donpenney (commented):
/lgtm

PR #145, opened 2022-04-05
Random suffix and smart truncation for talo created object names

jc-rh:
https://issues.redhat.com/browse/CNF-2563
https://bugzilla.redhat.com/show_bug.cgi?id=2057209

Change summary:

- Random suffix added to object names with smart truncation based on different length limits
- Mappings between full composite names and real object names saved in CGU status
- Composite names also set on objects as annotation
- Minor delay to before starting reconciliation to minimize chances of wasteful reconciliation on stale CGU / policy status. It also largely addresses [bug 2043301](https://bugzilla.redhat.com/show_bug.cgi?id=2043301)
- Self correction logic in status update based on the name annotation when name mapping in CGU status is inconsistent with the child objects
- UT update and kuttl test update


controllers/clustergroupupgrade_controller.go
```diff
@@ -52,6 +54,8 @@ type ClusterGroupUpgradeReconciler struct {
 	Recorder record.EventRecorder
 }
 
+const statusUpdateWaitInSeconds = 3
```
irinamihai, line 57: This should be moved to utils/constants.go

controllers/clustergroupupgrade_controller.go
```diff
@@ -87,6 +91,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	r.Log.Info("Start recociling CGU", "version", clusterGroupUpgrade.GetResourceVersion())
```
irinamihai, line 94: Typo: "reconciling"

controllers/clustergroupupgrade_controller.go
jc-rh: This is private to this file, similar to the ones in the other controller.

controllers/clustergroupupgrade_controller.go
jc-rh: Fixed

controllers/clustergroupupgrade_controller.go
```diff
@@ -164,6 +170,10 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					statusMessage = "The ClusterGroupUpgrade CR has all clusters already compliant with the specified managed policies"
 					nextReconcile = ctrl.Result{RequeueAfter: requeueAfter}
 				} else {
+					if clusterGroupUpgrade.Status.SafeResourceNames == nil {
+						clusterGroupUpgrade.Status.SafeResourceNames = make(map[string]string)
+					}
+					r.Log.Info("[Reconcile]", "name map", clusterGroupUpgrade.Status.SafeResourceNames)
```
irinamihai, line 176:
```suggestion
					r.Log.Info("[Reconcile]", "Status.SafeResourceNames", clusterGroupUpgrade.Status.SafeResourceNames)
```
or something similar. I don't think "name map" is explanatory enough.

controllers/clustergroupupgrade_controller.go
```diff
@@ -841,7 +869,14 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyNameForCopiedPo
 		}
 		// Update the metadata name
 		metadataContent := metadata.(map[string]interface{})
-		metadataContent["name"] = getResourceName(clusterGroupUpgrade, metadataContent["name"].(string))
+		name := utils.GetResourceName(clusterGroupUpgrade, metadataContent["name"].(string))
+		safeName, ok := clusterGroupUpgrade.Status.SafeResourceNames[name]
+		if !ok {
+			safeName = utils.GetSafeResourceName(name, utils.MaxPolicyNameLength, 0)
+			clusterGroupUpgrade.Status.SafeResourceNames[name] = safeName
+
```
irinamihai, line 877: This line is extra.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1007,40 +1053,43 @@ func (r *ClusterGroupUpgradeReconciler) ensureBatchPlacementRule(ctx context.Con
...
+			"name":      name,
 			"namespace": clusterGroupUpgrade.Namespace,
 			"labels": map[string]interface{}{
 				"app": "openshift-cluster-group-upgrades",
 				"openshift-cluster-group-upgrades/clusterGroupUpgrade": clusterGroupUpgrade.Name,
-				"openshift-cluster-group-upgrades/forPolicy":           prName,
+				"openshift-cluster-group-upgrades/forPolicy":           policyName,
+			},
+			"annotations": map[string]interface{}{
+				utils.ResourceNameAnnotation: desiredName,
```
irinamihai, line 1092: This is the name the user initially wanted to set up for the resource? If so, I think I would change "ResourceNameAnnotation" to something more intuitive, maybe even "DesiredName" or "InitialName" or something similar.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1182,25 +1238,28 @@ func (r *ClusterGroupUpgradeReconciler) ensureBatchPlacementBinding(
 }
 
 func (r *ClusterGroupUpgradeReconciler) newBatchPlacementBinding(clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade,
-	placementRuleName string, placementBindingName string) *unstructured.Unstructured {
+	policyName, placementRuleName, name, desiredName string) *unstructured.Unstructured {
```
irinamihai, line 1241:
```suggestion
	policyName, placementRuleName, placementBindingName, desiredName string) *unstructured.Unstructured {
```

controllers/clustergroupupgrade_controller.go
```diff
@@ -1501,9 +1559,19 @@ func (r *ClusterGroupUpgradeReconciler) buildRemediationPlan(
 	return nil
 }
 
-func getResourceName(clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, initialString string) string {
-	return clusterGroupUpgrade.Name + "-" + initialString
-}
+// func getGeneratedResourceNamePrefix(clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, initialString string) string {
```
irinamihai, line 1562: Hmm, do we need to keep this code since it's commented out?

controllers/utils/multicloud_util.go
```diff
@@ -136,14 +140,18 @@ var EnsureInstallPlanIsApproved = func(
 		if installPlan.Spec.Approved {
 			multiCloudLog.Info("InstallPlan has already been approved",
 				"InstallPlan", installPlan.ObjectMeta.Name, "namespace", installPlan.ObjectMeta.Namespace)
-			return InstallPlanCannotBeApproved, nil
+			return InstallPlanWasApproved, nil
```
irinamihai, line 143: This is still not accurate as TALO didn't create a managedClusterAction to approve the InstallPlan.
I think the best approach here would be to add another return value, InstallPlanAlreadyApproved and return that instead.

controllers/utils/multicloud_util.go
```diff
@@ -98,13 +98,17 @@ var EnsureInstallPlanIsApproved = func(
 		clusterGroupUpgrade, subscription.Status.InstallPlanRef.Kind, subscription.Status.InstallPlanRef.Name)
 	multiCloudLog.Info("[EnsureInstallPlanIsApproved] Create MCV for InstallPlan", "InstallPlan",
 		subscription.Status.InstallPlanRef.Name, "ns", clusterName)
+	safeName, ok := clusterGroupUpgrade.Status.SafeResourceNames[mcvForInstallPlanName]
```
irinamihai, line 101: These 3 lines are used throughout the code, though with different params for GetSafeResourceName. Wondering if we could make a function to hold these lines and return the safeName.

controllers/utils/common.go
```diff
@@ -26,3 +29,18 @@ func GetMinOf3(number1, number2, number3 int) int {
 		return number3
 	}
 }
+
+// GetSafeResourceName creates a safe name to use with random suffix and possible truncation based on limits passed in
+func GetSafeResourceName(name string, maxLength, spareLength int) string {
+	const randomLength = 5
```
irinamihai, line 35: This 5 isn't really random here  :grin:

controllers/clustergroupupgrade_controller.go
```diff
@@ -1770,8 +1871,23 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 	})
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&ranv1alpha1.ClusterGroupUpgrade{}).
-		Owns(placementRuleUnstructured).
+		For(&ranv1alpha1.ClusterGroupUpgrade{}).WithEventFilter(predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// This is a hack so we can ignore CGU status update only
```
irinamihai, line 1876: Maybe we need a Jira to track this issue and remove the hack one the issue is solved?

controllers/clustergroupupgrade_controller.go
```diff
@@ -1561,15 +1628,43 @@ func (r *ClusterGroupUpgradeReconciler) getAllClustersForUpgrade(ctx context.Con
 	return clusterNames, nil
 }
 
+func (r *ClusterGroupUpgradeReconciler) updateChildResourceNames(ctx context.Context, safeNameMap map[string]string, childResourceNameList []string, u *unstructured.Unstructured) ([]string, error) {
```
irinamihai, line 1631: I think we need a description for this function. Also, it's unclear what "childResource" means just from looking at the function. We're also not updating names in this function, we're returning a list of names. Maybe something like "GetUpdatedResourceNameList"?

I'm confused about a couple of things here:
1. Why are we deleting resources here... Is it if it happens to create multiple policies/placement rules/placement bindings for the same CGU, but with different names?
2. Given how this function is called, how do you know you are deleting the oldest created object? Does it even matter what object gets deleted?
3. It was my understanding that the ResourceNameAnnotation is keeping the desired name, but at line 1634 we are comparing it to the safe name.

Maybe it will all make more sense tomorrow morning :smiley:

controllers/clustergroupupgrade_controller.go
jc-rh: This was added while I was debugging the issue where reconcile calls work on stale CGU status. I meant to remove it during cleanup. Let me know if you prefer to keep it. BTW, I had added the resource version log later. It's easy to tell if it's happening or not from comparing that log line across multiple reconcile calls.

controllers/clustergroupupgrade_controller.go
jc-rh: removed

controllers/clustergroupupgrade_controller.go
jc-rh: Changed it DesiredResourceNameAnnotation

controllers/clustergroupupgrade_controller.go
jc-rh: Agreed

controllers/clustergroupupgrade_controller.go
jc-rh: Meant to remove this too

controllers/clustergroupupgrade_controller.go
jc-rh: The issue was raised 5 years ago and still no fix planned. Realistically, it probably will never get fixed. So I am not sure if adding a jira will help us.

controllers/clustergroupupgrade_controller.go
jc-rh: Yes, I can add a description. For your questions:
1. Yes, whenever we see multiple items from the query with the same desired name in their annotation, we need to correct the situation by deleting the extra, as they won't get updated if the managed policy content changes. Only the one being tracked by the name map should be kept.
2. This kind of conflict only happens when the reconcile call starts with a stale status/name map (empty or missing names) and creates new names (i.e. new resources) in the map. At the end, before updating status, this function is used like a sanity check. The ones that are already in the map just get created in this reconcile call.
3. No, u.GetName() is the safe name of the object. I guess I will change the variable name to desiredName.

controllers/utils/multicloud_util.go
jc-rh: Done

controllers/clustergroupupgrade_controller.go
irinamihai: I think we can keep it for now.

controllers/clustergroupupgrade_controller.go
irinamihai: Do we need to keep "Annotation" at the end? Seems extra.

tests/kuttl/tests/adjust-max-concurrency/00-assert.yaml
```diff
@@ -21,19 +21,33 @@ status:
     status: "False"
     type: Ready
   copiedPolicies:
-  - cgu-adjust-max-conc-policy1-common-cluster-version-policy
-  - cgu-adjust-max-conc-policy2-common-pao-sub-policy
+  - cgu-adjust-max-conc-policy1-common-cluster-versio-kuttl
```
irinamihai, line 24: Would this actually work? The length is already 56, combined with the namespace "default" we'd get "default.cgu-adjust-max-conc-policy1-common-cluster-versio-kuttl" which has a length of 64. Am I missing something here?

controllers/clustergroupupgrade_controller.go
```diff
@@ -1628,23 +1591,24 @@ func (r *ClusterGroupUpgradeReconciler) getAllClustersForUpgrade(ctx context.Con
 	return clusterNames, nil
 }
 
-func (r *ClusterGroupUpgradeReconciler) updateChildResourceNames(ctx context.Context, safeNameMap map[string]string, childResourceNameList []string, u *unstructured.Unstructured) ([]string, error) {
-	if name, ok := u.GetAnnotations()[utils.ResourceNameAnnotation]; ok {
-		if safeName, ok := safeNameMap[name]; ok {
+// checkDuplicateChildResources goes through the list of resources against the current names map. Duplicate (with same desired name annotation value) ones are removed.
```
irinamihai, line 1594: Please check the format for other function descriptions used throughout the file, for ex getNextNonCompliantPolicyForCluster

tests/kuttl/tests/adjust-max-concurrency/00-assert.yaml
jc-rh: Length is actually 55

controllers/clustergroupupgrade_controller.go
jc-rh: Done

serngawy (changes requested):
few comment and just to clarify;  you add the suffix either way if the original name is exceeded 63 chars or not ? also please squash the commits

controllers/clustergroupupgrade_controller.go
```diff
@@ -1630,18 +1690,22 @@ func (r *ClusterGroupUpgradeReconciler) updateStatus(ctx context.Context, cluste
...
 		err = r.Status().Update(ctx, clusterGroupUpgrade)
-
 		return err
 	})
 
 	if err != nil {
 		return err
 	}
+	// Wait a bit so that API server/etcd syncs up and the next reconsile call will get the updated CGU
+	time.Sleep(statusUpdateWaitInMilliSeconds * time.Millisecond)
```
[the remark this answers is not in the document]
jc-rh: Without this, we almost always get another reconcile call on a stale version of CGU right after one set of policies just get copied. We end up with two sets of copied policies and will rely on the name annotation based duplicate check for cleaning them up. Another side effect is that any reconcile call on a stale object will get flagged and requeued right away, which often cause another reconcile call on stale cgu again as we update status everytime. It can go on for several times until we are lucky enough to hit etcd node that just gets updated in the previous call. A bit wasteful in general and logs will look confusing. I can remove it if it really impacts our scale objective.

jc-rh:
> few comment and just to clarify; you add the suffix either way if the original name is exceeded 63 chars or not ? also please squash the commits

Yes, to be consistent, we always add the suffix, in the same spirit as mentioned in CNF-2563.
Yes, I will squash it. That's why this PR is still on hold. I'd like to keep the commits around until review is done.

controllers/clustergroupupgrade_controller.go
```diff
@@ -481,19 +487,23 @@ func (r *ClusterGroupUpgradeReconciler) remediateCurrentBatch(
 		if managedPolicyIndex == utils.AllPoliciesValidated || managedPolicyIndex == utils.NoPolicyIndex {
 			continue
 		}
+		prName := utils.GetResourceName(clusterGroupUpgrade, clusterGroupUpgrade.Status.ManagedPoliciesForUpgrade[managedPolicyIndex].Name+"-placement")
```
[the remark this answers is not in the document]
jc-rh: Will update.

controllers/utils/multicloud_util.go
```diff
@@ -219,18 +224,21 @@ func EnsureManagedClusterView(
 
 // EnsureManagedClusterActionForInstallPlan creates or updates an action for an InstallPlan.
 func EnsureManagedClusterActionForInstallPlan(
-	ctx context.Context, c client.Client, name string, namespace string,
+	ctx context.Context, c client.Client, safeName, name, namespace string,
```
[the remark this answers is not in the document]
jc-rh: "name" is actually used in the annotation. for logging though, we probably should use safeName, or it's gonna be a bit confusing.

jc-rh:
@serngawy @vitus133 This PR is ready as squash merge has been enabled in our prow config: https://github.com/openshift/release/pull/28328

PR #146, opened 2022-04-05
Ignore policies that have remediationAction enforce

serngawy:
Ignore policies with remediation Action enforce from being processed by CGU controller

Signed-off-by: melserngawy <melserng@redhat.com>


donpenney (commented):
/lgtm

PR #147, opened 2022-04-06
Use builtin client in createResourcesFromTemplates

vitus133:
Unit tests relay on the controller fake client. Using other clients make
code impossible to test.
This commit removes one of the instances where dynamic client was
used making it possible to unit-tests it.
In addition, it fixes a bug discovered in progress, where template
was rendered with uninitialized data.


rcarrillocruz:
Oh nevemind, @sabbir-47 had already lgtm'd .
This is good, removing tech debt.

PR #156, opened 2022-04-11
remove null fields for precaching from api response

sabbir-47:
- when preCaching is set to false, no precaching field is returned 
in the api response
- when preCaching is set to true, only non null fields are returned

/cc @vitus133 @rcarrillocruz @danielmellado @donpenney 


controllers/precache.go
```diff
@@ -29,6 +29,22 @@ func (r *ClusterGroupUpgradeReconciler) reconcilePrecaching(
...
+				Spec: &ranv1alpha1.PrecachingSpec{
+					PlatformImage:                "",
+					OperatorsIndexes:             []string{},
+					OperatorsPackagesAndChannels: []string{},
+				},
+				Status:   make(map[string]string),
+				Clusters: clusters,
+			}
+		}
+
```
vitus133, line 47: Can this block move after checking the `doneCondition`? Otherwise we'll be calling `getAllClustersForUpgrade` every time we reconcile, even after the precaching was already done. This is a heavy function

controllers/precache.go
```diff
@@ -38,7 +54,7 @@ func (r *ClusterGroupUpgradeReconciler) reconcilePrecaching(
 			return nil
 		}
 		// Precaching is required and not marked as done
-		return r.precachingFsm(ctx, clusterGroupUpgrade)
+		return r.precachingFsm(ctx, clusterGroupUpgrade.Status.Precaching.Clusters, clusterGroupUpgrade)
```
vitus133, line 57: Why are we passing both `clusterGroupUpgrade` and its inner structure?

vitus133 (changes requested):
Hi, got a couple of questions about the restructuring.
Thanks!

controllers/precache.go
sabbir-47: You are right, i can move to the getAllClustersForUpgrade in the original place at precacheFsm.go file, in that case i will keep `Clusters: []string{}` initially, and then set the Clusters field later. Because of using pointer, we have to initialize field before we can write over them.

sabbir-47: Fixed as suggested with 8af5d7ad847ce4a50b06e95e0923bcb71563855c

controllers/precache.go
sabbir-47: Fixed as suggested with https://github.com/openshift-kni/cluster-group-upgrades-operator/commit/8af5d7ad847ce4a50b06e95e0923bcb71563855c

config/manager/related-images/.gitignore
```diff
@@ -0,0 +1 @@
+patch.yaml
```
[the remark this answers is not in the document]
sabbir-47: Rebased, thanks

donpenney (approved):
/lgtm

PR #160, opened 2022-04-13
cleanup backup view and job if upgrade succeeds

sabbir-47:
- Refactored  precachingCleanup function to jobAndViewCleanup
- When upgrade succeeds jobAndViewCleanup function removes
both view and stale job for backup and precaching

/cc @vitus133 @rcarrillocruz @donpenney @danielmellado 

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -338,10 +338,10 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 
 				if isUpgradeComplete {
-					err = r.precachingCleanup(ctx, clusterGroupUpgrade)
+					err = r.jobAndViewCleanup(ctx, clusterGroupUpgrade)
 					if err != nil {
-						msg := fmt.Sprint("Precaching cleanup failed with error:", err)
-						r.Recorder.Event(clusterGroupUpgrade, corev1.EventTypeWarning, "PrecachingCleanupFailed", msg)
+						msg := fmt.Sprint("Job and managedclusterview cleanup failed with error:", err)
+						r.Recorder.Event(clusterGroupUpgrade, corev1.EventTypeWarning, "JobAndViewCleanupFailed", msg)
 					}
 					meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
 						Type:    "Ready",
--- a/controllers/managedClusterResources.go
+++ b/controllers/managedClusterResources.go
@@ -271,33 +271,53 @@ func (r *ClusterGroupUpgradeReconciler) renderYamlTemplate(
 
 // precachingCleanup deletes all precaching objects. Called when upgrade is done
 // returns: 			error
-func (r *ClusterGroupUpgradeReconciler) precachingCleanup(
+func (r *ClusterGroupUpgradeReconciler) jobAndViewCleanup(
 	ctx context.Context,
 	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) error {
 
-	if clusterGroupUpgrade.Spec.PreCaching {
-		clusters, err := r.getAllClustersForUpgrade(ctx, clusterGroupUpgrade)
-		if err != nil {
-			return fmt.Errorf("[precachingCleanup]cannot obtain the CGU cluster list: %s", err)
+	var viewTemplate, deleteTemplate []resourceTemplate
+
+	switch {
+	case clusterGroupUpgrade.Spec.Backup && clusterGroupUpgrade.Spec.PreCaching:
+		for _, v := range append(backupView, precacheAllViews...) {
+			viewTemplate = append(viewTemplate, v)
+		}
+		for _, v := range append(backupDeleteTemplates, precacheDeleteTemplates...) {
+			deleteTemplate = append(deleteTemplate, v)
 		}
 
-		for _, cluster := range clusters {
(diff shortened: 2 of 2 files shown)


jc-rh:
Should this change follow the same logic as:

```
func (r *ClusterGroupUpgradeReconciler) takeActionsAfterCompletion(
	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) error {

	actionsAfterCompletion := clusterGroupUpgrade.Spec.Actions.AfterCompletion
	// Add/delete cluster labels
	if actionsAfterCompletion.AddClusterLabels != nil || actionsAfterCompletion.DeleteClusterLabels != nil {
		clusters, err := r.getAllClustersForUpgrade(ctx, clusterGroupUpgrade)
		if err != nil {
			return err
		}
		labels := map[string]map[string]string{
			"add":    actionsAfterCompletion.AddClusterLabels,
			"delete": actionsAfterCompletion.DeleteClusterLabels,
		}

		if err := r.manageClusterLabels(ctx, clusters, labels); err != nil {
			return err
		}
	}

	// Cleanup resources
	if actionsAfterCompletion.DeleteObjects == nil || *actionsAfterCompletion.DeleteObjects {
		err := r.deleteResources(ctx, clusterGroupUpgrade)
		if err != nil {
			return err
		}
	}

	return nil
}
```

controllers/clustergroupupgrade_controller.go
```diff
@@ -338,10 +338,10 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 
 				if isUpgradeComplete {
-					err = r.precachingCleanup(ctx, clusterGroupUpgrade)
+					err = r.jobAndViewCleanup(ctx, clusterGroupUpgrade)
```
jc-rh, line 341: The current pattern is to update CGU status only in this cycle and cleanup happens in the next reconcile. If no particular reason why we have to do this as quickly as possible, maybe stick with the current pattern?

jc-rh: Always might be strong word. It's only introduced 6 days ago. It's fine if you want to deal with this in another PR. Not sure how we track things like this. Raise a JIRA?

jc-rh: One reason why you might want to move now is that the actionsAfterCompletion.DeleteObjects flag will still honored. Otherwise, you would have to add another if check?

jc-rh: Sorry, you are right. I read the history wrong.

vitus133: Can we split the discussion into two parts
1. actionsAfterCompletion.DeleteObjects flag
IIUC this flag is intended to delete policies, which are part of the API exposed to user.  However, managedclusterviews and spoke resources for pre-caching and backup are not part of the API. They are auxiliary and transient resources we don't ask user to create for us. IMHO, they must be always deleted, so this flag should not be checked for them.
1. cleanup that happens in the next reconcile for policies and on the current reconcile for the transient objects.
History-wise, both changes were introduced at about the same time, and weren't coordinated, so we need to choose which change is the pattern we want to adhere to, if any. Why postponing the resource deletion to the next reconcile, or why to hurry and delete the transient objects on this reconcile - are there pros and cons we should consider?

sabbir-47: I completely agree with @vitus133 on the point 1. For, internal and transient object  i.e. stale jobs, their namespaces and their associated manageclusterview cleanup must be done regardless to whether DeleteObjects has been set to true or false as this flag is only intend to be binded with the policy.

For 2nd point, i think this PR is not scoped to do it. We should discuss about it in a team meeting, and if decided, a JIRA issue can be raised to fix it, if required.

jc-rh: Thanks guys. Since I am new to this, everything looks new to me :smiley: Sorry, @sabbir-47 

To be honest, I don't have a strong preference on either points. The only thing I am against is inconsistency. The way it is going now, it feels like two code bases growing on their own in one repo. Let's discuss these two in the next standup or groom meeting. My thoughts right now:

1. I am not sure why it was introduced in the first place. @vitus133 to your point though, I would argue the same even for copied policies as well. Leaving them behind would be problematic if the source ones are updated. Maybe it was for special cases in dev environment only? If we decide to keep the flag, I think it should control everything.
2. From what I just learned in troubleshooting scale test, splitting the cleanup logic into the next reconcile call has some advantages. In general, it's desired for controllers/operators to be more responsive, so it can serve multiple CRs "at the same time". Finishing one CR a bit sooner is a lesser concern. Each reconcile call shouldn't take too long and try to do everything in one shot, considering our single threaded design. The price we pay is slightly higher resource usage on ACM by keeping the CRs for a little bit longer.

vitus133: Thanks for the explanation @jc-rh !
I created stories for the both points above:
1. Clarify the use case of the flag with whoever requested it: https://issues.redhat.com/browse/CNF-5012
2. Make the cleanup consistent: https://issues.redhat.com/browse/CNF-5013
Right now I think we should unblock @sabbir-47  from making the recovery functionality operational.
WDYT?

sabbir-47: Would be happy to get any review on the code i submitted, otherwise looking for 1 or 2 lgtm before i go on a vacation.

vitus133: /lgtm
Stories are there - we'll do it after the holidays :)

PR #161, opened 2022-04-13
Update operator bundle icon

nishant-parekh:
Using the Redhat logo for the time being
Making the linter happy! 

Signed-off-by: Nishant Parekh <nparekh@redhat.com>


vitus133 (approved):
Yes, this is indeed the Red Hat

vitus133:
/hold
this change should be done also in config/manifests/bases/cluster-group-upgrades-operator.clusterserviceversion.yaml
The file you've changed is autogenerated upon `make bundle`, so it will be overwritten

PR #164, opened 2022-04-14
Add example for recovery partition creation

donpenney:
Updated the recovery README.md with an example extra-manifest
MachineConfig that creates and mounts the recovery partition.

### DIFF
--- a/recovery/README.md
+++ b/recovery/README.md
@@ -167,6 +167,53 @@ spec:
                 name: backup
 ```
 
+## Create Recovery Partition
+
+The recovery partition can be created at install time by defining an extra-manifest MachineConfig, as described here:<br>
+<https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/README.md#deploying-a-site>
+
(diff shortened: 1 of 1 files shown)


recovery/README.md
```diff
@@ -167,6 +167,53 @@ spec:
...
+
+The following extra-manifest MachineConfig will create a 50G partition on the specified disk:
+
+```
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 98-var-recovery-partition
```
sabbir-47, line 183: Should we rather propose a generic name?

donpenney: I was trying to follow MachineConfig conventions. Is there something you'd prefer to see? I don't think the name impacts anything

sabbir-47: No it does not impact anything. Just wondering what would 98 mean, you can ignore it

vitus133: Hangs here long enough for  people to see :)
/lgtm

PR #175, opened 2022-04-26
Make image tags consistent with upstream builds

vitus133:
This makes image tags of the deployment and related images consistent
with the tags applied in quay to the images pushed by the upstream CI.
This is needed for installing the upstream built operator.
If this is not done, the operator installation will fail because of
image pull error.
In addition it reverts the operator-sdk version to the latest
one recommended by the OCP documentation (operator-sdk-v1.16.0-ocp).
The latter fixes 'make bundle' failures happening with operator-sdk 1.19.1

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cluster-group-upgrades-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.16.0-ocp
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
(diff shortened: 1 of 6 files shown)


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cluster-group-upgrades-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.16.0-ocp
```
jc-rh, line 9: @nishant-parekh is this the same version you had in your recent PR?

vitus133: /test ci/prow/images

vitus133: @nishant-parekh changed v1.16.0-ocp to 1.19.1 while annotating for the scorecard tests. I think it was an unintentional change (the annotations were intentional, but the version change - not). Please correct me if I'm wrong.
I think we need to have a fixed sdk version per branch. There is a makefile target (make operator-sdk) that downloads the right version.
If we don't make the sdk version fixed, we might have unintentional changes related to auto-generation with every PR.

PR #176, opened 2022-04-28
Bug 2079838: Make image tag consistent with u/s build

vitus133:
This makes image tags of the deployment and related images consistent
with the tags applied in quay to the images pushed by the upstream CI.
This is needed for installing the upstream built operator.
If this is not done, the operator installation will fail because of
image pull error.

/cc @danielmellado @donpenney  @jc-rh  @nishant-parekh @rcarrillocruz @sabbir-47

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.3
+VERSION ?= 4.10.0
 
 # You can use podman or docker as a container engine. Notice that there are some options that might be only valid for one of them.
 ENGINE ?= docker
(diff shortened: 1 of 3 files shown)


rcarrillocruz:
Seems the lgtm has been caught by the bot. Besides, It seems @vitus133 needs to prepend a BZ to the PR

vitus133:
> Besides, It seems @vitus133 needs to prepend a BZ to the PR

Let's wait for the integration test to finish. I hope it will merge without a bug :)

danielmellado:
@vitus133 I don't think that'll happen, sadly. Check the CI reply above:

> @vitus133: No Bugzilla bug is referenced in the title of this pull request.
To reference a bug, add 'Bug XXX:' to the title of this pull request and request another bug refresh with /bugzilla refresh.

This would block any merge that doesn't meet these requirements.

vitus133:
> This would block any merge that doesn't meet these requirements.

OK, I am on it. It seems that lgtm label now disappeared

rcarrillocruz:
I think the lgtm was never picked cos Daniel put a text after it (yes, bot shenanigans).

vitus133:
yay

PR #188, opened 2022-05-10
 Bug 2083631 Reset batch start time on batch timeout

jc-rh:
It's getting reset in the batch complete case only right now. This PR added the missing line for the timeout case.


imiller0 (approved):
lgtm

PR #195, opened 2022-05-14
Improve batch progress state

sudomakeinstall2:
Now show a meaningful message instead of numbers to indicate the
progress of the batch


serngawy (changes requested):
few comment and please make sure you have a full integration test with all cases. Thanks

controllers/clustergroupupgrade_controller.go
```diff
@@ -1157,33 +1161,25 @@ func (r *ClusterGroupUpgradeReconciler) newBatchPlacementRule(clusterGroupUpgrad
...
 	currentPolicyIndex := startIndex
-
-	for {
+	for ; currentPolicyIndex < numberOfPolicies; currentPolicyIndex++ {
 		// Get the name of the managed policy matching the current index.
 		currentManagedPolicyInfo := utils.GetManagedPolicyForUpgradeByIndex(currentPolicyIndex, clusterGroupUpgrade)
 		currentManagedPolicy, err := r.getPolicyByName(ctx, currentManagedPolicyInfo.Name, currentManagedPolicyInfo.Namespace)
 		if err != nil {
-			return utils.NoPolicyIndex, err
+			return -1, err
```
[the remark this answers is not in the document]
sudomakeinstall2: Before this PR the policy index for each cluster was stored in the status, this index could have been -1 when the progress has not been started and -2 when the progress was completed. This was confusing and not meaningful for the user. This PR removes the need for -1 and -2 for the index by introducing a state variable. So `utils.NoPolicyIndex` is removed. These return statements are error statements and the first value is not used and is not meaningful and I don't think there is a need to define a constant for it.

sudomakeinstall2:
> 

Does it need more tests? The field in the status that was changed is already in 6 integration tests that check `Completed` and `InProgress` states. `NotStarted` state is not really used and can't be observed because it immediately changes to `InProgress`.

serngawy:
> > 
> 
> Does it need more tests? The field in the status that was changed is already in 6 integration tests that check `Completed` and `InProgress` states. `NotStarted` state is not really used and can't be observed because it immediately changes to `InProgress`.
  
Fine then, I wasn't aware of the integration test checks

controllers/clustergroupupgrade_controller.go
```diff
@@ -1197,33 +1193,22 @@ func (r *ClusterGroupUpgradeReconciler) getNextNonCompliantPolicyForCluster(
...
-		nextNonCompliantPolicyIndex, err := r.getNextNonCompliantPolicyForCluster(ctx, clusterGroupUpgrade, batchClusterName, currentPolicyIndex)
-		if err != nil {
-			return false, err
-		}
-
-		// If we can find a managed policy for which the cluster is NonCompliant, it means the upgrade has
-		// not finished.
-		if nextNonCompliantPolicyIndex != utils.NoPolicyIndex {
-			return false, nil
+		clusterProgress := clusterGroupUpgrade.Status.Status.CurrentBatchRemediationProgress[batchClusterName]
```
jc-rh, line 1205: We can chat about this. The point of the function is to check again periodically after it reaches the timeout state. Obviously it wasn't exactly working due to the bug you fixed a few lines above. However, it is mentioned in our readme. If we want to change the behavior, we should create a JIRA and do it in another PR.

vitus133 (commented):
Just a minor comment, otherwise looks good

controllers/clustergroupupgrade_controller.go
sudomakeinstall2: @jc-rh  I agree but I don't understand how the behaviour is changed. It seems to me that the function wants to check whether all the clusters in the current batch are completed. I think the changes are consistent with the past behaviour. It checks every cluster status to be completed.

jc-rh: Maybe I missed something here? The original code actually checks the policy compliant status again. The new code only looks at the policy index map?

sudomakeinstall2: you are right I missed it. I will fix it.

sudomakeinstall2: @jc-rh I updated this part, although I am pretty sure that this part of the code doesn't work as expected currently.

vitus133 (approved):
Great!

PR #200, opened 2022-05-27
Bug 2087768:  Verify backup disk size before triggering backup task in the spoke

sabbir-47:
- Added a new file recovery/backup_size.go
- It verifies the estimated backup size against the free disk space

/cc @donpenney @danielmellado @vitus133 @jc-rh 


recovery/cmd/backup_size.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+var (
+	clusterPath         = "/var/lib/etcd/member/snap/db"
+	staticPodsPath      = "/etc/kubernetes/static-pod-resources/"
+	usrLocalPath        = "/usr/local"
+	kubeletPath         = "/var/lib/kubelet"
+	etcPath             = "/etc"
+	binPath             = "/etc/kubernetes/static-pod-resources/bin/"
+	backupSizeSafetyNet = 10.00
```
donpenney, line 20: Maybe add units in a comment, or in the var name? Also, just wondering are these variables configurable, or should these be const?

recovery/cmd/backup_size.go
```diff
@@ -0,0 +1,189 @@
...
+		if units[i] == ebytes {
+			ebytesID = i
+		}
+		if units[i] == fbytes {
+			fbytesID = i
+		}
+	}
+
+	if ebytesID <= fbytesID {
+		// safetynet of 10GiB is added
```
donpenney, line 65: Comment could get stale by having the hardcoded value, which a reader could lookup. Or if this is configurable, could be inaccurate. I'd just remove the value and not that the size requirement is being padded

recovery/cmd/backup_size.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var diskPartition []disk.PartitionStat
+
+	diskPartition, err := disk.Partitions(false)
+	if err != nil {
+		log.Println(err)
+	}
+
+	var totalDiskSize, freeDiskSpace float64
+	for _, v := range diskPartition {
```
donpenney, line 125: Are you totalling all partitions in this loop? I'd have expected we're just looking at /var/recovery

sabbir-47: If the extra partition is not created and /var/recovery is not a separate partition then what should we do?

recovery/cmd/backup_size.go
sabbir-47: I kept a a value for now which is hardcoded to 10GiB which can be changed based on our discussion but i was thinking if there is a case that after taking the backup there is no sufficient space left for the upgrade to continue, what will we do? The reason being introducing the safetyNet is to be sure we have some space left after the backup is taken.

donpenney: I'm just suggesting you drop "10GiB" from the comment here, as it could easily get outdated

recovery/cmd/backup_size.go
donpenney: We'd just care about the free space available under /var/recovery, whether it's a separate partition or not

recovery/cmd/backup_size.go
sabbir-47: Added

recovery/cmd/backup_size.go
sabbir-47: Dropped with commit effab4c9023ada38a70df52005e59d4bda2a8714

recovery/cmd/backup_size.go
sabbir-47: changed as suggested with commit effab4c9023ada38a70df52005e59d4bda2a8714

recovery/cmd/backup_size.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	freeDiskSpace = float64(stat.Bavail * uint64(stat.Bsize))
+
+	err = unix.Statfs("/", &stat)
+	if err != nil {
+		return freeDiskSpace, err
+	}
+	total, tbytes := sizeConversion(float64(stat.Blocks * uint64(stat.Bsize)))
+	log.Infof("Total disk size is: %.2f %s", total, tbytes)
```
donpenney, line 138: I assume this is here for informational purposes only? The size of "/" wouldn't be the "Total disk size", though - it's just the total size of the rootfs, which would be irrelevant to /var/recovery. Either way, I'm not sure why we'd need to log the total disk size

recovery/cmd/backup_size.go
```diff
@@ -0,0 +1,188 @@
...
+		if freeDisk > backupSizeSafetyNet+estimated {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// estimateDiskSize calculate the required backup size
+// returns: disk size(int), error
+func estimateDiskSize() (float64, error) {
```
donpenney, line 79: Suggest renaming to something like estimateFsSpaceRequirements. The current function name sounds like it's getting the size of the disk.

recovery/cmd/backup_size.go
sabbir-47: yes, "/" is considered as rootfs, in any case, it is excluding "/boot" mountpoint since we can't use that. This is only for information purpose, it is not needed to calculate or find the free and/or estimated space.

donpenney: My point, though, is that this would not be relevant to /var/recovery (particularly if either /var/recovery or /var is a separate partition), and is not "Total disk size". It is just the size of the root filesystem itself, not the disk

sabbir-47: Removed with 74f7f6eca77c0b69ab1530c7ac44eb2c204a8f8b

PR #206, opened 2022-06-01
Prevents reconcile error when policy status is missing

sudomakeinstall2:
Handles the error when the policy status is missing and reconcile in 30 seconds


sudomakeinstall2:
@jc-rh Can you take a look at this?

controllers/utils/common.go
```diff
@@ -58,3 +58,10 @@ func NewSafeResourceName(name, suffix string, maxLength, spareLength int) string
 	}
 	return fmt.Sprintf("%s-%s", base, suffix)
 }
+
+// PolicyStatusUnknownError defines an error when policy status is missing or unknown
+type PolicyStatusUnknownError string
```
jc-rh, line 63: I like this idea of having an error type with a message. In this case though, it might be a bit of an overkill. I would be ok with logging the different cases only before returning. Maybe just add this as another constant along with the existing ones in constants.go. The more interesting case is actually the non compliant case where the non compliant message could be useful for troubleshooting, if the policy just doesn't go compliant for a cluster. We will revisit this as part of https://issues.redhat.com/browse/CNF-3903.

controllers/clustergroupupgrade_controller.go
```diff
@@ -174,13 +174,21 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent, err :=
 					r.doManagedPoliciesExist(ctx, clusterGroupUpgrade, true)
 				if err != nil {
+					if _, ok := err.(utils.PolicyStatusUnknownError); ok {
```
jc-rh, line 177: Not sure if we need to have a specific requeue after value for this case, as we can just treat it as non-compliant. Why not just do that in getClusterComplianceWithPolicy and getClustersNonCompliantWithPolicy. We already do that for the case where the map is there but the cluster entry doesn't exist.

PR #209, opened 2022-06-07
Bug 2084962: Remove hard limit of 100 concurrent cluster upgrades

fontivan:
- Removed the hard limit of upgrading 100 clusters concurrently


fontivan:
I have uploaded a test image to http://quay.io/rockwerk/cluster-group-upgrades-operator:bz-2084962
Perhaps we can test this in the scale lab @jc-rh / @ijolliffe?

PR #217, opened 2022-06-15
ART-4052: Add requirements for ART bundle creation

thiagoalessio:
since the presence of `image-references` in the bundle dir was messing up with CI, I suggest we just create a copy in `manifests/` with the structure that makes ART's automation happy, while keeping `bundle/` untouched.


donpenney (commented):
/lgtm

nishant-parekh (approved):

PR #226, opened 2022-06-21
Enhancement CNF-5294: Improve batch timeout calculation to use the remaining time for calculating the remaining timeout

fontivan:
For each batch the timeout will be calculated using the remaining time and the remaining number of batches. This will cause the timeout to continuously adjust itself for batches that complete quickly or slowly early on by giving the remaining batches more or less time respectively.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -328,23 +328,44 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					}
 
 					// Check for batch timeout.
-					batchTimeout := time.Duration(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout/len(clusterGroupUpgrade.Status.RemediationPlan)) * time.Minute
-					if !clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.IsZero() && time.Since(clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.Time) > batchTimeout {
-						if len(clusterGroupUpgrade.Spec.RemediationStrategy.Canaries) != 0 &&
-							clusterGroupUpgrade.Status.Status.CurrentBatch <= len(clusterGroupUpgrade.Spec.RemediationStrategy.Canaries) {
-							r.Log.Info("Canaries batch timed out")
-							meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
-								Type:    "Ready",
-								Status:  metav1.ConditionFalse,
-								Reason:  "UpgradeTimedOut",
-								Message: "The ClusterGroupUpgrade CR policies are taking too long to complete",
-							})
-						} else {
-							r.Log.Info("Batch upgrade timed out")
-							clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt = metav1.Time{}
-							if clusterGroupUpgrade.Status.Status.CurrentBatch < len(clusterGroupUpgrade.Status.RemediationPlan) {
-								clusterGroupUpgrade.Status.Status.CurrentBatch++
+					remainingTime := float64(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout) - time.Since(clusterGroupUpgrade.Status.Status.StartedAt.Time).Seconds()
+					// Because the length of the plan will include the current batch but the status index starts at 0 we need to add one to the remaining batches
+					remainingBatches := len(clusterGroupUpgrade.Status.RemediationPlan) - clusterGroupUpgrade.Status.Status.CurrentBatch + 1
+
+					// Must avoid a division by 0 here in case this is the last batch
+					if remainingBatches > 0 {
+
+						// The current batch's timeout shall be the remaining time divided by the number of batches remaining
+						// This is to ensure we are giving each batch as much time as possible within the remaining allotment
+						currentBatchTimeout := time.Duration(remainingTime / float64(remainingBatches) * float64(time.Second))
+
+						// Check if this batch has timed out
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -328,25 +328,46 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
-								Status:  metav1.ConditionFalse,
-								Reason:  "UpgradeTimedOut",
-								Message: "The ClusterGroupUpgrade CR policies are taking too long to complete",
-							})
-						} else {
-							r.Log.Info("Batch upgrade timed out")
-							clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt = metav1.Time{}
-							if clusterGroupUpgrade.Status.Status.CurrentBatch < len(clusterGroupUpgrade.Status.RemediationPlan) {
-								clusterGroupUpgrade.Status.Status.CurrentBatch++
+					remainingTime := float64(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout) - time.Since(clusterGroupUpgrade.Status.Status.StartedAt.Time).Seconds()
```
jc-rh, line 331: Keep re-calculating this would eventually use up all the remaining time on the current batch if it never completes? I think this should only be done once per batch.

controllers/clustergroupupgrade_controller.go
```diff
@@ -328,25 +328,46 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
-								Status:  metav1.ConditionFalse,
-								Reason:  "UpgradeTimedOut",
-								Message: "The ClusterGroupUpgrade CR policies are taking too long to complete",
-							})
-						} else {
-							r.Log.Info("Batch upgrade timed out")
-							clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt = metav1.Time{}
-							if clusterGroupUpgrade.Status.Status.CurrentBatch < len(clusterGroupUpgrade.Status.RemediationPlan) {
-								clusterGroupUpgrade.Status.Status.CurrentBatch++
+					remainingTime := float64(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout) - time.Since(clusterGroupUpgrade.Status.Status.StartedAt.Time).Seconds() - time.Since(clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.Time).Seconds()
```
jc-rh, line 331: The since calls confuse me a bit. I think it's right mathmatically but it could probably be simplified.

jc-rh: How about batchStartTime.Sub(cguStartTime)

controllers/clustergroupupgrade_controller.go
```diff
@@ -327,26 +327,48 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+							} else {
+								// This was not a canary but we already know we timed out
+								// Setting the remaining time to be negative here will trigger the default case below
+								remainingTime = -1
 							}
 						}
 					}
+
+					// If there is zero or negative time remaining then we have timed out
+					if remainingTime <= 0 {
```
jc-rh, line 365: Now this case can go back to the else block above.

controllers/clustergroupupgrade_controller.go
```diff
@@ -327,9 +327,23 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 						nextReconcile = ctrl.Result{RequeueAfter: requeueAfter}
 					}
 
-					// Check for batch timeout.
-					batchTimeout := time.Duration(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout/len(clusterGroupUpgrade.Status.RemediationPlan)) * time.Minute
-					if !clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.IsZero() && time.Since(clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.Time) > batchTimeout {
+					// The remaining time will be the total timeout subtract the elapsed time spent on both the batch and cgu
+					// It is important to include the current batch here so that we don't let the entire timeout be consumed by a single batch that gets stuck
+					remainingTime := float64(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout) - clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.Time.Sub(clusterGroupUpgrade.Status.Status.StartedAt.Time).Seconds()
```
serngawy, line 332: better to check for nil value before clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.Time

serngawy, line 332: is the clusterGroupUpgrade.Spec.RemediationStrategy.Timeout always in sec ? better to create time.Duration and then subtract the seconds

serngawy (changes requested):

controllers/clustergroupupgrade_controller.go
fontivan: I can reorganize the if checks to make sure this is checked for zero before it is used

fontivan: As far as I can tell this is always in seconds, but it is just an integer. It is also treated as seconds elsewhere. I would still have to assume this is seconds to make a duration no?

serngawy:
Sounds good to me , just to confirm did u test it end to end ? like having a case when the timeout correctly calculated and works ?

fontivan:
> Sounds good to me , just to confirm did u test it end to end ? like having a case when the timeout correctly calculated and works ?

Yes I was testing this last week and this morning. The current batch timeout looks good for multiple runs

serngawy (approved):

PR #227, opened 2022-06-22
Enhancement CNF-5013: Delete precaching objects on final cleanup

fontivan:
Makes the clean up more consistent

### DIFF
--- a/controllers/actions.go
+++ b/controllers/actions.go
@@ -164,6 +164,11 @@ func (r *ClusterGroupUpgradeReconciler) deleteResources(
 	}
 	clusterGroupUpgrade.Status.CopiedPolicies = nil
 
+	err = r.jobAndViewCleanup(ctx, clusterGroupUpgrade)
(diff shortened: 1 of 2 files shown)


jc-rh:
Why did we keep the existing call?

sabbir-47:
> Why did we keep the existing call?

I think the initial idea was to clean the backup and/or precaching job and its related managedclusterView objects just before update the operator status on upgrade has been complete. In any case, have you verified with the delete resource actually does the delete of these jobs and objects as expected?

fontivan:
> > Why did we keep the existing call?
> 
> I think the initial idea was to clean the backup and/or precaching job and its related managedclusterView objects just before update the operator status on upgrade has been complete. In any case, have you verified with the delete resource actually does the delete of these jobs and objects as expected?

I checked this today by performing an upgrade with backup enabled, watched the backup run, and can confirm there are no MCV objects present after it completes

PR #242, opened 2022-07-04
Bug 2090885: Fix treating missing clusters in policy compliance status as compliant

sudomakeinstall2:
Now checks specifically if the cluster is compliant with policy and don't consider missing clusters as compliant.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -758,7 +758,7 @@ func (r *ClusterGroupUpgradeReconciler) doManagedPoliciesExist(
 			if filterNonCompliantPolicies {
 				// Check the policy has at least one of the clusters from the CR in NonCompliant state.
 				clustersNonCompliantWithPolicy, err := r.getClustersNonCompliantWithPolicy(
-					ctx, clusterGroupUpgrade, foundPolicy, true)
+					ctx, clusterGroupUpgrade, foundPolicy)
 				if err != nil {
 					return false, nil, nil, err
 				}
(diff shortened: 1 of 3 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -1421,45 +1421,25 @@ func (r *ClusterGroupUpgradeReconciler) getPolicyClusterStatus(policy *unstructu
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot obtain all the details about the clusters in the CR: %s", err)
 	}
-
-	// Filter only the clusters present in the current upgrade.
-	if filterClustersForUpgrade {
-		var nonCompliantClustersInUpgrade []string
-		allClustersForUpgrade, err := r.getAllClustersForUpgrade(ctx, clusterGroupUpgrade)
+	for _, cluster := range allClustersForUpgrade {
+		compliance, err := r.getClusterComplianceWithPolicy(cluster, policy)
```
jc-rh, line 1433: we can't redo this for every single cluster. it's the same api call.

sudomakeinstall2: That makes sense however I kinda repeated what `getClustersNonCompliantWithManagedPolicies` does :laughing:

sudomakeinstall2: I am on it as soon as it passes the tests

jc-rh: Don't worry about it in this PR then. Let's get this fix in first.

jc-rh:
It seems to be a conscious decision to treat missing cluster as compliant. We need to revisit the reasons behind that decision before changing this behavior.

jc-rh (approved):

PR #240, opened 2022-06-30
Bug 2102743: RBAC permissions are not granted for ServiceMonitor CR

rauhersu:
Permissons added for the monitoring group when making requests
to the API server for the actions specified under role.yaml

PR #248, opened 2022-07-06
Revert "BZ 2102743: RBAC permissions are not granted for ServiceMoni…

rauhersu:
…tor CR (#240)"

This reverts commit b22846e46d74ca8947a8174c9ff663a8df7ac929.

The reason to revert this PR is because `role.yaml` was edited manually to set the permissions. We are going to set the permissions automatically using an `operator-sdk` tag which will set the permissions on role.yaml on running `make bundle` as part of a posterior PR.


donpenney (approved):
/lgtm

PR #249, opened 2022-07-06
Add check of bundle generated files to ci-job

donpenney:
Added hack/check-git-tree.sh utility to check git tree for unstaged or
untracked files, and updated the ci-job target to run this check after
generating bundle files. This will ensure generated file changes will
not be missed by commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ bashate: ## Run bashate
 	hack/bashate.sh
 
 .PHONY: ci-job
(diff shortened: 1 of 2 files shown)


donpenney:
Note: This ci-job test failure is expected, as currently there are some stale generated files. Once the PRs to fix these files have merged, this test should then succeed.

/cc @jc-rh @nishant-parekh

nishant-parekh (approved):

PR #251, opened 2022-07-07
Bug:2102743 RBAC permissions are not granted for ServiceMonitor CR ...

rauhersu:
Permissions for metrics are granted on `role.yaml` using automatic configuration generation on introducing the
corresponding` //+kubebuilder:rbac` tag on `clustergroupupgrade_controller.go`


donpenney (commented):
/lgtm

PR #253, opened 2022-07-07
Add retry to golangci-lint installation

donpenney:
The installation of golangci-lint has been seen to fail occasionally
due to github rate limiting. Added a retry to the golangci-lint.sh to
install golangci-lint via "go install" method if standard installation
script method fails.

/cc @jc-rh


jc-rh (approved):

donpenney:
The golangci-lint task doesn't exist in 4.10, so that cherry-pick fails and isn't needed.

PR #255, opened 2022-07-11
Bug 2105992: Fix batch timeout calculation units (and add unit tests)

fontivan:
The reference timeout units were incorrectly referred to as seconds instead of as minutes in the solution for CNF-5294. This fixes that, and also adds a number of unit tests that validate the new behaviour performs as intended.


jc-rh (changes requested):
Maybe move this function to utils?

fontivan:
> Maybe move this function to utils?

Completed in latest commit

jc-rh (approved):

PR #257, opened 2022-07-11
Bug 2087125: Check all batches for upgrade complete

jc-rh:
Last batch processing now evaluates all batches before declaring upgrade complete.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -369,7 +369,12 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					}
 				} else {
 					// On last batch, check all batches
-					if r.isUpgradeComplete(ctx, clusterGroupUpgrade) {
+					var isUpgradeComplete bool
+					isUpgradeComplete, err = r.isUpgradeComplete(ctx, clusterGroupUpgrade)
+					if err != nil {
+						return
+					}
+					if isUpgradeComplete {
 						meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
 							Type:    "Ready",
 							Status:  metav1.ConditionTrue,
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -315,86 +313,82 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					} else {
+						// Add the needed cluster names to upgrade to the appropriate placement rule.
+						nextReconcile, err = r.remediateCurrentBatch(ctx, clusterGroupUpgrade)
+						if err != nil {
+							return
+						}
+
+						// Check if this batch has timed out
+						if !clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.IsZero() {
+
```
donpenney, line 341: This line has trailing whitespace (tabs to indent, but it's a blank line so unneeded). The "go fmt" step that's done by ci-job then updates the file to remove this whitespace, which gets caught by the bundle-check (which looks for a "dirty" working tree after running the ci-job build steps)

controllers/clustergroupupgrade_controller.go
```diff
@@ -315,86 +313,83 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					// On last batch, check all batches
+					if r.isUpgradeComplete(ctx, clusterGroupUpgrade) {
+						meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
+							Type:    "Ready",
+							Status:  metav1.ConditionTrue,
+							Reason:  "UpgradeCompleted",
+							Message: "The ClusterGroupUpgrade CR has all clusters compliant with all the managed policies",
+						})
+						nextReconcile = requeueImmediately()
+					} else if !clusterGroupUpgrade.Status.Status.StartedAt.IsZero() && time.Since(clusterGroupUpgrade.Status.Status.StartedAt.Time) > time.Duration(clusterGroupUpgrade.Spec.RemediationStrategy.Timeout)*time.Minute {
```
serngawy, line 380: better to use time.Before() instead of doing the conversation. check this 

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Time.Before

controllers/clustergroupupgrade_controller.go
```diff
@@ -1224,27 +1219,35 @@ func (r *ClusterGroupUpgradeReconciler) getNextNonCompliantPolicyForCluster(
...
+				clusterProgress := clusterGroupUpgrade.Status.Status.CurrentBatchRemediationProgress[batchClusterName]
+				if clusterProgress.State == ranv1alpha1.Completed {
+					continue
+				}
+				startIndex := 0
+				if clusterProgress.PolicyIndex != nil {
+					startIndex = *clusterProgress.PolicyIndex
+				}
+				nextNonCompliantPolicyIndex, err := r.getNextNonCompliantPolicyForCluster(ctx, clusterGroupUpgrade, batchClusterName, startIndex)
+				if err != nil {
```
serngawy, line 1240: better syntax 

				if err != nil || nextNonCompliantPolicyIndex < len(clusterGroupUpgrade.Status.ManagedPoliciesForUpgrade)  {
					return false
				}

serngawy (commented):
sounds good , just few comments

controllers/clustergroupupgrade_controller.go
jc-rh: Not sure how this function is better than Since()

controllers/clustergroupupgrade_controller.go
jc-rh: Good point. Err should have been handled differently.

serngawy (approved):

PR #262, opened 2022-07-15
Enhancement CNF-5558: Add new method for specifying clusters for CGU

fontivan:
* Added clusterLabelSelectors field to the CGU spec
* Marked clusterSelectors field as deprecated
* Updated method that determines what clusters are included in CGU to account for the above changes
* Added `debug` option to make to allow connecting debugger via IDE plugins
* Added debug binary to .gitignore

### DIFF
--- a/api/v1alpha1/clustergroupupgrade_types.go
+++ b/api/v1alpha1/clustergroupupgrade_types.go
@@ -113,8 +113,29 @@ type ClusterGroupUpgradeSpec struct {
 	//   - label1Name
 	// All the clusters matching the labels specified in clusterSelector will be included
 	// in the update plan.
+	// Deprecated: Use ClusterLabelSelectors instead
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Cluster Selector",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	ClusterSelector []string `json:"clusterSelector,omitempty"`
+	// This field holds a list of expressions or labels that will be used to determine what clusters to include in the operation.
+	// The expected format is as follows:
+	// clusterLabelSelectors:
+	//   - matchExpressions:
+	//       - key: label1
+	//         operator: In
+	//         values:
+	//           - value1a
+	//           - value1b
+	//   - matchLabels:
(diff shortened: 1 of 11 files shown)


api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -113,8 +113,29 @@ type ClusterGroupUpgradeSpec struct {
...
+	//           - value1a
+	//           - value1b
+	//   - matchLabels:
+	//       label2: value2
+	//   - matchExpressions:
+	//       - key: label3
+	//         operator: In
+	//         values:
+	//           - value3
+	//     matchLabels:
```
imiller0, line 135: Is this "matchLabels" missing a leading "-", eg:
```suggestion
	//    - matchLabels:
```

imiller0 (commented):
Are there unit tests that should be updated, or added, with this change?

api/v1alpha1/clustergroupupgrade_types.go
fontivan: No, this third one is just an example of a label selector that includes both expression and label elements, compared to the first two which are respectively just an expression and just a label

api/v1alpha1/clustergroupupgrade_types.go
```diff
@@ -113,8 +113,29 @@ type ClusterGroupUpgradeSpec struct {
 	//   - label1Name
 	// All the clusters matching the labels specified in clusterSelector will be included
 	// in the update plan.
+	// Deprecated: Use ClusterLabelSelectors instead
```
jc-rh, line 116: This is just a comment, right? Is there a markup for kubebuilder so the API is generated with some deprecated annotation?

fontivan: I've added a kubebuilder annotation for deprevated version that I found in the docs here: https://book.kubebuilder.io/reference/markers/crd.html

jc-rh: :+1: , how did I miss the very next line ...

jc-rh (approved):

PR #265, opened 2022-07-19
Bug 2106101: Shorter requeue interval on approving installplan

jc-rh:
This makes TALM check the subscription status more often when waiting for the installplan to be created.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -555,7 +555,8 @@ func (r *ClusterGroupUpgradeReconciler) updatePlacementRules(ctx context.Context
 
 func (r *ClusterGroupUpgradeReconciler) approveInstallPlan(
 	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) (bool, error) {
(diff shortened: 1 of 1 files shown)


donpenney:
A brief description of how this addresses the issue, in the commit message and/or PR description, would be helpful here

controllers/clustergroupupgrade_controller.go
```diff
@@ -555,7 +555,8 @@ func (r *ClusterGroupUpgradeReconciler) updatePlacementRules(ctx context.Context
 
 func (r *ClusterGroupUpgradeReconciler) approveInstallPlan(
 	ctx context.Context, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade) (bool, error) {
-	multiCloudPendingStatus := false
+
+	reconcileSooner := false
```
sabbir-47, line 559: Why do we change boolean var name? it doesn't change any behavior for the function. Irrespective to the name, if the return value is true, the reconcile happen in shorter interval. The fix for Bug 2087125 also included this fix, no?

donpenney (commented):
/lgtm

PR #266, opened 2022-07-20
Bug 2077974: Prevent reconcile error when policy status is missing

sudomakeinstall2:
Consider the the cluster non compliant when the policy has missing status


jc-rh (commented):
Does it make more sense to change the getPolicyClusterStatus function instead? Less change to these caller functions. Also, there could be real error from that function that now is being shadowed?

sudomakeinstall2:
How do you want to change that function? Its job is to return the cluster status and it is doing its job. The bug is not in that function. Also in my opinion the changes to the caller functions is good because I think the extraneous error handling is removed.

PR #276, opened 2022-08-03
Bug 2109961: Run before enable actions even if all clusters are already compliant with cgu policy

fontivan:
* Ensure that the before enable actions are run even if all clusters are already compliant
    * Update status for all clusters already being compliant so that the cgu will finalize and run after completion actions
    * Do not requeue after the cgu is completed

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -214,6 +214,12 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 
 				if allManagedPoliciesExist {
+					// Trigger the before enable actions now since they should run even if all the clusters are already compliant.
+					err = r.takeActionsBeforeEnable(ctx, clusterGroupUpgrade)
+					if err != nil {
+						return
+					}
+
 					// Build the upgrade batches.
 					err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
 					if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -222,10 +228,12 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 
 					// Set default values for status reason and message.
 					var statusReason, statusMessage string
+					var statusCondition metav1.ConditionStatus = metav1.ConditionFalse
```
jc-rh, line 231: ```statusCondition := metav1.ConditionFalse```

controllers/clustergroupupgrade_controller.go
```diff
@@ -214,6 +214,12 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 
 				if allManagedPoliciesExist {
+					// Trigger the before enable actions now since they should run even if all the clusters are already compliant.
+					err = r.takeActionsBeforeEnable(ctx, clusterGroupUpgrade)
```
jc-rh, line 218: The name of this field is misleading. It should have been something like `actionsOnEnable` or `actionsAfterEnable`, e.g. the ztp-running label was meant to show that policy remediation for ztp is in progress. So we can't do it here.

controllers/clustergroupupgrade_controller.go
```diff
@@ -435,9 +437,9 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 				// Set completion time only after post actions are executed with no errors
 				clusterGroupUpgrade.Status.Status.CompletedAt = metav1.Now()
+				nextReconcile = doNotRequeue()
```
jc-rh, line 440: It should be doNotRequeue already. Are you trying to get rid of excessive reconcile calls post completion?

controllers/clustergroupupgrade_controller.go
jc-rh: Now that I think about it, the `plan == nil` check should have been done later anyways. The blocking CR checks should still be done even if there is nothing to do for this CGU.

jc-rh (approved):

PR #277, opened 2022-08-04
Enhancement CNF-3932: Configurable action on batch timeout

fontivan:
* Allow specifying what to do when a batch timeout occurs
* Available options include:
    - Continue
    - Abort
* The default option if it is not specified is Continue
* May be expanded in the future with additional options

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -385,7 +385,6 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 											Reason:  "UpgradeTimedOut",
 											Message: "The ClusterGroupUpgrade CR policies are taking too long to complete",
 										})
-									case ranv1alpha1.BatchTimeoutAction.Continue:
 									default:
 										// If the value was continue or not defined then continue
 										clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt = metav1.Time{}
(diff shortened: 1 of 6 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -376,9 +376,22 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+									switch clusterGroupUpgrade.Spec.BatchTimeoutAction {
+									case ranv1alpha1.BatchTimeoutAction.Abort:
+										// If the value was abort then we need to fail out
+										meta.SetStatusCondition(&clusterGroupUpgrade.Status.Conditions, metav1.Condition{
+											Type:    "Ready",
+											Status:  metav1.ConditionFalse,
+											Reason:  "UpgradeTimedOut",
+											Message: "The ClusterGroupUpgrade CR policies are taking too long to complete",
+										})
+									case ranv1alpha1.BatchTimeoutAction.Continue:
```
sabbir-47, line 388: What you are going to do in this case ranv1alpha1.BatchTimeoutAction.Continue: ? if nothing is set or don, shouldn't the default case suffice?

fontivan: The default case would suffice but I just like being a bit explicit. I can just use the default case if you want though

sabbir-47: I would not keep a case if we are not performing any action on that case.

fontivan: already updated it in the latest commit

jc-rh (approved):

PR #285, opened 2022-08-08
Bug 2115480: Fix batch timeout calculation for first batch

fontivan:
The current batch index is forced to be indexed from one instead of zero, so when passiing it into the timeout calculation function it needs to be subtracted again. This is consistent with how the batch index is handled around remediation policies.


controllers/clustergroupupgrade_controller.go
```diff
@@ -360,13 +360,17 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 						// Check if this batch has timed out
 						if !clusterGroupUpgrade.Status.Status.CurrentBatchStartedAt.IsZero() {
 
+							// Because the current batch index automatically advances from 0 to 1 on the first loop
+							// We need to subtract one when passing it in here to ensure the math is correct
 							currentBatchTimeout := utils.CalculateBatchTimeout(
 								clusterGroupUpgrade.Spec.RemediationStrategy.Timeout,
 								len(clusterGroupUpgrade.Status.RemediationPlan),
-								clusterGroupUpgrade.Status.Status.CurrentBatch,
+								clusterGroupUpgrade.Status.Status.CurrentBatch-1,
```
jc-rh, line 368: I know this way the unit test doesn't need to change. It's easier to read though if this is done inside the utility function and the test data in unit test matches what we expect.

serngawy (approved):

PR #288, opened 2022-08-10
Bug 2075585: Support for hub cluster template function fromConfigMap

Missxiaoguo:
ACM hub cluster resource-specific template functions are restricted to the same
namespace as the policy is defined on the hub cluster due to security concerns,
so TALM cannot work with hub cluster template as TALM creates the copied policies
in the namespace where the CGU is created. This commit updates to support the
hub-side resource-specific function fromConfigMap by replicating the configmap
to the CGU namespace.

Changes included:
 - validate managed policies before reconciling resources, requeue with medium interval
   and update CGU status when invalid policies found
 - search for hub templates in each managed policy and replicate configmap in the CGU namespace
 - update the policy hub template with replicated configmap
 - requeue with medium interval when unsupported templates found
   i.e. printf variable in resource Name/Namespace field
        fromSecret/lookup functions
 - unittest for valid and invalid hub templates usage

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -843,7 +843,7 @@ func (r *ClusterGroupUpgradeReconciler) doManagedPoliciesExist(
 
 			// If the parent policy is not valid due to missing field, add its name to the list of invalid policies.
 			policyErr := utils.VerifyPolicyObjects(foundPolicy)
-			if policyErr != nil || policyInvalidHubTmpl[managedPolicyName] {
+			if policyErr != nil {
 				r.Log.Error(policyErr, "Policy is invalid")
 				managedPoliciesInfo.invalidPolicies = append(managedPoliciesInfo.invalidPolicies, managedPolicyName)
 				continue
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -56,6 +56,12 @@ type ClusterGroupUpgradeReconciler struct {
 	Recorder record.EventRecorder
 }
 
+type policiesInfo struct {
+	invalidPolicies []string
+	missingPolicies []string
+	presentPolicies []*unstructured.Unstructured
+}
+
 const statusUpdateWaitInMilliSeconds = 100
 
 func doNotRequeue() ctrl.Result {
@@ -91,7 +97,7 @@ func requeueWithCustomInterval(interval time.Duration) ctrl.Result {
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=action.open-cluster-management.io,resources=managedclusteractions,verbs=create;update;delete;get;list;watch;patch
 //+kubebuilder:rbac:groups=view.open-cluster-management.io,resources=managedclusterviews,verbs=create;update;delete;get;list;watch;patch
-//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules,verbs=get;list;watch;create;update;patch;delete
@@ -206,9 +212,8 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			} else if readyCondition.Reason == "UpgradeNotStarted" || readyCondition.Reason == utils.CannotStart {
 				// Before starting the upgrade check that all the managed policies exist.
 				var allManagedPoliciesExist bool
-				var managedPoliciesMissing []string
-				var managedPoliciesPresent []*unstructured.Unstructured
-				allManagedPoliciesExist, managedPoliciesMissing, managedPoliciesPresent, err =
+				var managedPoliciesInfo policiesInfo
+				allManagedPoliciesExist, managedPoliciesInfo, err =
 					r.doManagedPoliciesExist(ctx, clusterGroupUpgrade, true)
 				if err != nil {
 					return
@@ -216,7 +221,7 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 
 				if allManagedPoliciesExist {
 					// Build the upgrade batches.
-					err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesPresent)
+					err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesInfo.presentPolicies)
 					if err != nil {
 						return
 					}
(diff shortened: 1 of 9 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -935,35 +981,73 @@ func (r *ClusterGroupUpgradeReconciler) copyManagedInformPolicy(
...
-	// Go through the template array.
-	for _, template := range policyTemplatesArr {
-		// Get to the metadata name of the ConfigurationPolicy.
-		objectDefinition := template.(map[string]interface{})["objectDefinition"]
-		if objectDefinition == nil {
-			return fmt.Errorf("policy %s is missing its spec.policy-templates.objectDefinition", managedPolicyName)
+func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyHubTemplate(
+	ctx context.Context, objectDef interface{}, cguNamespace, managedPolicyName, managedPolicyNamespace string) error {
+	// Process only if the managed policy is not created in the CGU namespace
+	if managedPolicyNamespace == cguNamespace {
```
jc-rh, line 1015: is this an actual case we hit sometimes?

controllers/clustergroupupgrade_controller.go
```diff
@@ -1018,33 +1102,21 @@ func (r *ClusterGroupUpgradeReconciler) getPolicyContent(
 	for _, template := range policyTemplatesArr {
 		// Get to the metadata name of the ConfigurationPolicy.
 		objectDefinition := template.(map[string]interface{})["objectDefinition"]
-		if objectDefinition == nil {
```
jc-rh, line 1021: I see that we currently only call this with already validated policies. Maybe add some comment for the function so that it doesn't get reused with newly loaded policies that are not validated yet.

controllers/clustergroupupgrade_controller.go
```diff
@@ -1395,27 +1467,35 @@ func (r *ClusterGroupUpgradeReconciler) getCopiedPolicies(ctx context.Context, c
...
 		if err != nil {
-			return err
+			if _, ok := err.(*utils.PolicyErr); ok {
+				// If it's a policy error(i.e. unsupported hub template),
+				// break the loop to execute updateChildResourceNamesInStatus
+				// to update the CGU status with already created policies
+				isPolicyErr = true
+				break
+			}
+			return isPolicyErr, err
```
jc-rh, line 1484: Would it more clear if we just do `return false, err`? Same for the next two returns

controllers/clustergroupupgrade_controller.go
Missxiaoguo: Sure, will update

Missxiaoguo: done

controllers/clustergroupupgrade_controller.go
Missxiaoguo: done

controllers/clustergroupupgrade_controller.go
jc-rh: ah I see.

controllers/clustergroupupgrade_controller.go
```diff
@@ -765,14 +784,25 @@ func (r *ClusterGroupUpgradeReconciler) doManagedPoliciesExist(
...
 		}
 
+		for _, policyT := range childPolicy.Spec.PolicyTemplates {
+			configPolicy := &unstructured.Unstructured{}
+			json.Unmarshal(policyT.ObjectDefinition.Raw, configPolicy)
+
+			configPlcTAnnotations := configPolicy.GetAnnotations()
+			// Identify policies have invalid hub templates
+			_, ok := configPlcTAnnotations["policy.open-cluster-management.io/hub-templates-error"]
+			if ok || strings.Contains(string(policyT.ObjectDefinition.Raw), "{{hub") {
```
imiller0, line 794: Is the string '{{hub...' part of the value from the policy.open-cluster-management.io/hub-templates-error annotation? If so it may make sense to restrict the string search to just the value rather than the whole config policy.

controllers/clustergroupupgrade_controller.go
```diff
@@ -791,20 +821,35 @@ func (r *ClusterGroupUpgradeReconciler) doManagedPoliciesExist(
...
+			// If the parent policy has invalid hub template, add its name to the list of invalid policies.
+			if policyInvalidHubTmpl[managedPolicyName] {
+				r.Log.Error(&utils.PolicyErr{ObjName: managedPolicyName, ErrMsg: utils.PlcHasHubTmplErr}, "Policy is invalid")
+				managedPoliciesInfo.invalidPolicies = append(managedPoliciesInfo.invalidPolicies, managedPolicyName)
+				continue
+			}
+
+			// If the parent policy is not valid due to missing field, add its name to the list of invalid policies.
+			policyErr := utils.VerifyPolicyObjects(foundPolicy)
+			if policyErr != nil || policyInvalidHubTmpl[managedPolicyName] {
```
imiller0, line 841: The test for policyInvalidHubTmpl here is redundant with the test on line 833. Is there a reason it was included here as well?

controllers/utils/policy_template_test.go
```diff
@@ -0,0 +1,160 @@
+package utils
```
imiller0, line 1: Really nice unit testing!

controllers/clustergroupupgrade_controller.go
Missxiaoguo: Good catch! Shouldn't check on the whole config policy as the annotation has "{{hub". Will update

controllers/clustergroupupgrade_controller.go
Missxiaoguo: Good catch! That was added by mistake.  Will update

Missxiaoguo: Done

Missxiaoguo:
rebase

controllers/clustergroupupgrade_controller.go
imiller0: Ah, I missed that you are checking the child policy. Is the intent here that if the _child_ policy contains "{{hub" then ACM failed to do the template substitution and the child policy is thus invalid?
If so, is this a test that is periodically re-evaluated?

Missxiaoguo: Yes, if the child policy contains "{{hub", it means the hub template processing failed on the hub and the root policy is invalid. The hub templating is processed on the hub when policies are created or updated during propogation of repliciated policies to spoke clusters. The hub-template-error annotation should be added to the child policy by the propogator on the hub if there has invalid template. In case the annotation was not added,  the config policy controller running on the spoke cluster also checks if child policy has '{{hub' and adds the annotation, so we are unlikely to have the situation where the policy contains invalid template without the error annotation attached. Just in case something weird happens, I check the presence of '{{hub' against child policy here as well. 

It's re-evaluated if CGU is not ready to be remediated due to any reason, ie. not all resources have been prepared, Cgu is not enabled etc

imiller0: I wonder if it would be more efficient to simply look for the string every time? In the current implementation we are unmarshalling the entire policy every time to do this check and then, if there is not an error (no annotation) we scan the entire policy for a string comparison.

controllers/utils/policy_template.go
```diff
@@ -0,0 +1,183 @@
...
+	err := r.Get(ctx, fromResource, cm)
+	if err != nil {
+		return err
+	}
+
+	copiedCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        toResource.Name,
+			Namespace:   toResource.Namespace,
+			Annotations: cm.GetAnnotations(),
```
imiller0, line 151: We should include here an indication that TALM copied the configmap. If I remember correctly multiple policies that refer to the same configmap will make only one copy (is that right?) and all of them may overwrite/update it at any time, so we can't actually have it "owned" by one CGU, but we should be able to at least note that TALM created it.

Missxiaoguo: Yes, we only have one copy for all policies. I added these labels to the copied CM. Do you think it's good enough? 

openshift-cluster-group-upgrades/fromCmName
openshift-cluster-group-upgrades/fromCmNamespace

controllers/clustergroupupgrade_controller.go
Missxiaoguo: Good point. Yep, we actually just need check the string {{hub. Will update.

controllers/utils/policy_template.go
imiller0: When we copy policies we annotate with:
`cluster-group-upgrades-operator/rname: .....`
For consistency I would say we should use the same prefix and combine into a single value (makes changes by multiple CGU atomic):
`cluster-group-upgrades-operator/fromCmName: namespace.name`

Missxiaoguo: Updated.

imiller0 (approved):
/lgtm

PR #291, opened 2022-08-11
[release-4.10] Enhancement CNF-3932: Configurable action on batch timeout (Cherry Pick)

fontivan:
* Enhancement [CNF-3932](https://issues.redhat.com//browse/CNF-3932): Configurable action on batch timeout
* Allow specifying what to do when a batch timeout occurs
* Available options include:
    - Continue
    - Abort
* The default option if it is not specified is Continue
* May be expanded in the future with additional options
* Fix comment to meet linter requirements
* Remove explicit continue & just use default


fontivan:
After discussion we have decided not to backport this to 4.10 since it is a new feature. Will be available in 4.11 instead

PR #293, opened 2022-08-15
Enhancement CNF-4565: Do not create empty fields for managed policies that do not have appropriate content

fontivan:
* Only create the managed policies field for policies that meet the general requirements for policy content, namely;
  1. Have kind subscription
  2. Have a namespace
  3. Have a name
* This will prevent policies from appearing with a content of `null` in the cgu status

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -848,6 +848,12 @@ func (r *ClusterGroupUpgradeReconciler) processManagedPolicyForUpgradeContent(
 			return err
 		}
 
+		// If the policy types did not return with useful data then continue to the next policy
+		// We don't want to create empty fields in the map with null values as that is confusing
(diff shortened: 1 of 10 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -848,6 +848,12 @@ func (r *ClusterGroupUpgradeReconciler) processManagedPolicyForUpgradeContent(
 			return err
 		}
 
+		// If the policy types did not return with useful data then continue to the next policy
+		// We don't want to create empty fields in the map with null values as that is confusing
+		if policyTypes == nil || len(policyTypes) == 0 {
```
jc-rh, line 853: Minor comment. len() in go is actually nil safe: https://go.dev/tour/moretypes/12

jc-rh (approved):

jc-rh (approved):

PR #224, opened 2022-06-21
Requeue immediately when moving to the next batch

jc-rh:

PR #296, opened 2022-08-16
Bug 2115480: Reconcile immediately after timeout

fontivan:
* After timing out on a batch we should reconcile immediately
* If we don't we will just waste 5 minutes of timeout waiting for the next loop


jc-rh (approved):
Thanks, I guess I missed this in #224.

PR #302, opened 2022-08-23
add skipRange for OLM upgrade support

nishant-parekh:
Signed-off-by: Nishant Parekh <nparekh@redhat.com>


config/manifests/bases/cluster-group-upgrades-operator.clusterserviceversion.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
+    olm.skipRange: '>=4.3.0-0 <4.12.0'
```
jc-rh, line 7: why 4.3.0 rather than 4.10.0?

jc-rh (approved):

PR #314, opened 2022-09-07
CNF-6019: Improve reconcile loop conditions/reasons/messages

fontivan:
* This is a vertical slice focused on improving the conditions/reasons/messages for TALM specifically in the reconcile loop
    * This is a continuation of the previous vertical slice

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -1803,11 +1803,25 @@ func (r *ClusterGroupUpgradeReconciler) blockingCRsNotCompleted(ctx context.Cont
 			continue
 		}
 
-		// If we find a blocking CR with a status different than "UpgradeCompleted", then we add it to the list.
+		// If we find a blocking CR that does not contain a successful completion condition then we add it to the list.
+		// For example, we need a condition like this:
+		/*
+			- message: The ClusterGroupUpgrade CR has all clusters compliant with all the managed policies
+			  reason: UpgradeCompleted
+			  status: "True"
+			  type: Succeeded
+		*/
+		upgradeCompleted := false
 		for i := range cgu.Status.Conditions {
-			if cgu.Status.Conditions[i].Reason != "UpgradeCompleted" {
-				blockingCRsNotCompleted = append(blockingCRsNotCompleted, cgu.Name)
+			if cgu.Status.Conditions[i].Reason == string(utils.ConditionReasons.UpgradeCompleted) &&
+				cgu.Status.Conditions[i].Status == "True" &&
+				cgu.Status.Conditions[i].Type == string(utils.ConditionTypes.Succeeded) {
+				upgradeCompleted = true
+
 			}
+		}
+		if !upgradeCompleted {
+			blockingCRsNotCompleted = append(blockingCRsNotCompleted, cgu.Name)
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


fontivan:
Having difficulties getting integration tests to run locally so wanted to see wht they look like on the CI

controllers/clustergroupupgrade_controller.go
```diff
@@ -195,283 +256,290 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					err = r.remediateCurrentBatch(ctx, clusterGroupUpgrade, &nextReconcile)
 					if err != nil {
 						return
 					}
 
-					if isBatchComplete {
-						// If the upgrade is completed for the current batch, cleanup and move to the next.
-						r.Log.Info("[Reconcile] Upgrade completed for batch", "batchIndex", clusterGroupUpgrade.Status.Status.CurrentBatch)
-						r.cleanupPlacementRules(ctx, clusterGroupUpgrade)
-						r.addClustersStatusOnCompleteBatch(clusterGroupUpgrade)
```
jc-rh, line 368: @sudomakeinstall2 this is still needed right?

controllers/clustergroupupgrade_controller.go
```diff
@@ -178,6 +178,53 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}
 
+	progressingCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.Progressing))
+
+	// Before starting the upgrade check that all the managed policies exist
+	var allManagedPoliciesExist bool
+	var managedPoliciesInfo policiesInfo
+
+	// doManagedPoliciesExist should only be called before the CGU has begun processing
+	if progressingCondition != nil {
```
jc-rh: `managedPoliciesInfo` needs to be populated for `buildRemediationPlan` and `processManagedPolicyForUpgradeContent` to work. So `doManagedPoliciesExist` can't be executed once only. It is a mandatory part of the process. In a way you could say the validated condition doesn't mean much as we always have to redo it until the cgu is started.

fontivan: @sabbir-47 perhaps we could change the proposal to move Validated to be after precaching and backup? Otherwise it seems we would need to rearchitect quite a bit of our reconcile loop

controllers/clustergroupupgrade_controller.go
```diff
@@ -194,282 +213,332 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
 
-				if allManagedPoliciesExist {
-					// Build the upgrade batches.
-					err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesInfo.presentPolicies)
-					if err != nil {
-						return
-					}
+		}
+	} else if (backupCondition == nil || (backupCondition != nil && (backupCondition.Reason == string(utils.ConditionReasons.Completed) || backupCondition.Reason == string(utils.ConditionReasons.PartiallyDone)))) &&
```
jc-rh, line 250: Is it better to add the condition with a reason like "not enabled" so that this expression can be a bit simpler?

jc-rh: Or at least wrap this up in a bool function with an easy-to-read name.

PR #317, opened 2022-09-09
CNF-6120:Update operator-framework module

rauhersu:
Updating operator-framework api to 0.10.4 (see go.mod). This is  the latest operator framework we can use without significant API changes on other dependencies.

The rest of the changes have been automatically generated on running these commands:

$ make common-deps-update
$ go mod vendor


danielmellado:
Will this be for OCP 4.11? If so, it'd make also sense to update go to 1.18

jc-rh:
0.10.4 seems to be still older than what OLM uses. I don't see the installplan field that deprecates installplanref in this version.

rauhersu:
After discussing with @jc-rh, we noticed the operator api update and the refactor of `InstallPlanRef` were orthogonal tasks -  the `SubscriptionStatus` has not  deprecated the former type even on the main api branch -.

We  have created two different Jira stories (which will end up in two different PRs) to manage this: on the one hand, the api [update](https://issues.redhat.com/browse/CNF-6120) and on the other hand, the `InstallPlanRef` [refactor](https://issues.redhat.com/browse/CNF-6202). 

I am therefore closing this PR which contains all the changes for both stories.

PR #332, opened 2022-09-22
CNF-5558: Improve cluster selection when combining multiple methods of specifing clusters in a CGU

fontivan:
* After discussing with Jun yesterday, it is not a super great idea to sort the entire list
* If clusters are manually specified in the CGU, then the user probably wants them done in the order they specified
* Clusters chosen via selectors / label selectors will be sorted alphabetically and processed after any explicitly listed clusters


jc-rh:
Maybe reverse the order of the clusters in one or two kuttl tests?

jc-rh (approved):

PR #333, opened 2022-09-22
Move validation up before pre-caching/backup

jc-rh:
Also added cgu patch commands to speed up integration test.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -177,6 +177,73 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}
 
+	progressingCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.Progressing))
+
+	// Set the progressing condition if it is not present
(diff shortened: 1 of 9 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -247,296 +310,227 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
-					if err != nil {
-						return
-					}
-				}
 			}
 		}
-
 	}
 
 	// Update status
```
sabbir-47, line 536: We need to handle the error here.

jc-rh: Error from where?

sabbir-47: Sorry, the comment is not related to this PR, so i couldn't add the comment in line 543. It is in line 543:

	err = r.updateStatus(ctx, clusterGroupUpgrade)
	return

jc-rh: Ok. The err is returned. It's not an error we should be handling.

sabbir-47: you are right.

PR #335, opened 2022-09-26
Cnf 6147 initiate backup when enable is true

sabbir-47:
initiate backup when enable is true
    - precaching is done before backup
    - backup is initiated when enabled is true
    - if backup is true, remediation plan is rebuild again

/cc @jc-rh @fontivan @vitus133 @alosadagrande


controllers/clustergroupupgrade_controller.go
```diff
@@ -248,152 +291,132 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
 
-				// Build the upgrade batches.
-				err = r.buildRemediationPlan(ctx, clusterGroupUpgrade, managedPoliciesInfo.presentPolicies)
-				if err != nil {
+		backupCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.BackupSuceeded))
+		if backupCondition != nil && backupCondition.Status == metav1.ConditionFalse {
+			utils.SetStatusCondition(
+				&clusterGroupUpgrade.Status.Conditions,
+				utils.ConditionTypes.Progressing,
+				utils.ConditionReasons.Enabled,
```
sabbir-47, line 321: Hi @jc-rh, @fontivan, when backup is going on, although although enabled is true, progressing condition waits for backup to succeed, in that case, what could be an appropriate utils.ConditionReasons? For now, i kept it is enabled, which doesn't make a lot of sense. may be "NotStarted"? wdyt?

controllers/actions.go
```diff
@@ -171,6 +171,7 @@ func (r *ClusterGroupUpgradeReconciler) deleteResources(
 		return fmt.Errorf("failed to delete precaching objects for CGU %s: %v", clusterGroupUpgrade.Name, err)
 	}
 
+	// comment: should we cleanup in failed clusters as well?
```
sabbir-47, line 174: @jc-rh @vitus133, do we only delete resources where precaching and backup were successful, leaving the failed on for debugging or de we cleanup the objects in all the clusters?

fontivan: I belive we had determined before that leaving such resources do not actually aid in debugging

controllers/clustergroupupgrade_controller.go
fontivan: I'm not sure you need to set a condition here. There will already be a BackupSucceeded/InProgress condition present while the backup is ongoing. Couldn't you just remove the check for backup condition and handle it inside the if condition?

fontivan:
It is somewhat difficult to read this review since it also includes @jc-rh's commits from their open PR (#333)

controllers/clustergroupupgrade_controller.go
sabbir-47: " Couldn't you just remove the check for backup condition and handle it inside the if condition?" --> are you suggesting it to move at line 330?

fontivan: oh ok I see what you mean now. Yeah this is probably fine then

README.md
```diff
@@ -30,36 +30,40 @@ The ClusterGroupUpgrade controller is designed to perform an upgrade to a group
...
+  | | True | PartiallyDone | Precaching failed for x clusters | 
+  | | False | InProgress | Precaching is not completed | 
+  | | False | InProgress | Precaching is in progress for x clusters | 
+  | | False | Failed | Precaching failed for all clusters |
+  `BackupSucceeded` | True | Completed | Backup is completed for all clusters|
+  | | True | PartiallyDone | Backup failed for x clusters |
+   | | False | NotEnabled | Backup will start when enable is true |
+  | | False | InProgress | Backup is in progress for x clusters|
+  | | False | Failed | Backup failed for all the clusters |
+  `Progressing`| True | InProgress| The ClusterGroupUpgrade CR has upgrade policies that are still non compliant|
```
jc-rh, line 59: This message was for the old `UpgradeNotCompleted` state. Should we update it with something more consistent to `Progressing`?

README.md
```diff
@@ -30,36 +30,40 @@ The ClusterGroupUpgrade controller is designed to perform an upgrade to a group
...
+  | | False | Completed | The ClusterGroupUpgrade CR has all clusters compliant with all the managed policies |
+  | | False | NotStarted | The Cluster backup is in progress |
```
jc-rh, line 61: In Steven's PR (@fontivan), we are streamlining the cluster selection. We stop re-doing cluster selection/policy validation when progressing becomes true. With this change, the condition for that part would become more complicated. Do you think it would make more sense to consider backup part of progressing (as we can't redo cluster selection at that point)?

sabbir-47: The problem is if precaching and backup are set to true, and if couple of clusters fails, we have filter out the successful clusters and rebuild/update cluster selection and remediation plan. I don't know why https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/348 is marked as a bug, for me it is more of an improvement. But the easier way is to if Progressing is set to true (it is only true when there is no blocking CR, missing CR, precaching and backup is done or failed), we can just check progressing == true and don't do cluster selection/policy validation.

controllers/clustergroupupgrade_controller.go
```diff
@@ -361,43 +341,87 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+				}
 			}
-		}
 
+			backupCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.BackupSuceeded))
+			if clusterGroupUpgrade.Status.Backup == nil || (backupCondition != nil && backupCondition.Status == metav1.ConditionTrue) {
+
+				// rebuild remediation plan, since backup has finished.
+				if clusterGroupUpgrade.Status.Backup != nil {
+					var allManagedPoliciesExist bool
```
jc-rh, line 374: We could avoid this duplicated block if we requeue after backup is done.

controllers/clustergroupupgrade_controller.go
```diff
@@ -361,43 +341,87 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+						err = r.updateStatus(ctx, clusterGroupUpgrade)
+						nextReconcile = requeueWithShortInterval()
+						return
+					}
+				}
 			}
-		}
 
+			backupCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.BackupSuceeded))
+			if clusterGroupUpgrade.Status.Backup == nil || (backupCondition != nil && backupCondition.Status == metav1.ConditionTrue) {
```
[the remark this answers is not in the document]
sabbir-47: - **when backup is running it will be false:**
    **- false || (true && false)**
- when backup is false it will be true:
    - true || (false && false)
- when backup is done it will be true:
    - true || (true && true)

controllers/clustergroupupgrade_controller.go
sabbir-47: added a check [44b6fc1](https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/335/commits/44b6fc11b1d841dcb22459ed81178cc674de4774), it should avoid the block after progressing is set to true

controllers/clustergroupupgrade_controller.go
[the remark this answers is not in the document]
jc-rh: when backup is running, don't we always return from line 364?

controllers/backup.go
```diff
@@ -41,8 +41,12 @@ func (r *ClusterGroupUpgradeReconciler) reconcileBackup(
 			// Backup is done
 			return nil
 		}
+
 		// Backup is required and not marked as done
-		return r.triggerBackup(ctx, clusterGroupUpgrade)
+		if *clusterGroupUpgrade.Spec.Enable {
```
jc-rh, line 46: this is also redundant now that reconcileBackup is only called after enable and blocking CR check.

controllers/clustergroupupgrade_controller.go
jc-rh: this is not what I meant and it seems to break integration tests. you can put it back and I will merge it as soon as tests pass. we will address the comments in a followup PR.

controllers/clustergroupupgrade_controller.go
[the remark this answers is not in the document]
sabbir-47: Yeah, that is on purpose, when backup is progressing nothing else should happen in the SNO, we should always return from that point. When backup is done, failed or in error, only then we move to next block. This part remained the same as it was

README.md
sabbir-47: This is what is returned for now as a message. @fontivan any idea what message could be appropriate?

jc-rh (approved):

PR #348, opened 2022-09-29
OCPBUGS-1692: Prevent repeated cluster selection during reconcile and deletion

fontivan:
* Generate cluster list once and use it repeatedly
* This will prevent us from re-selecting clusters 5+ times during reconcile


controllers/clustergroupupgrade_controller.go
```diff
@@ -151,21 +151,25 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			utils.ConditionTypes.ClustersSelected,
+			utils.ConditionReasons.NotFound,
+			metav1.ConditionFalse,
+			fmt.Sprintf("Unable to select clusters: %s", err),
+		)
+		nextReconcile = requeueWithLongInterval()
+		err = r.updateStatus(ctx, clusterGroupUpgrade)
+		return
+	}
 	if !meta.IsStatusConditionTrue(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.ClustersSelected)) {
```
jc-rh, line 172: :+1: I was wondering if it's gonna be an infinite loop

controllers/clustergroupupgrade_controller.go
```diff
@@ -151,21 +151,25 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 
 	// Verify that the selected clusters are all valid
 	// If this has already been done we will skip it on further reconcile loops
+
```
jc-rh, line 154: Shouldn't this whole block be in the block under the progressing condition check?

fontivan: No, since I am using validateCR to generate the initial cluster list and using it again below I need it to run every time

jc-rh: That block is run every time until progressing becomes true. We don't want to re-select clusters if it's already in progress.

controllers/clustergroupupgrade_controller.go
```diff
@@ -149,11 +149,19 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	// Set the progressing condition if it is not present
+	if progressingCondition == nil || (progressingCondition.Status == metav1.ConditionFalse && progressingCondition.Reason != string(utils.ConditionReasons.Completed)) {
+
 		var reconcile bool
-		reconcile, err = r.validateCR(ctx, clusterGroupUpgrade)
+		if len(clusters) == 0 {
+			clusters, reconcile, err = r.validateCR(ctx, clusterGroupUpgrade)
+		} else {
+			_, reconcile, err = r.validateCR(ctx, clusterGroupUpgrade)
```
jc-rh, line 163: This means we don't overwrite it as soon as the plan has been populated?

fontivan: If the clusters list is defined here then it was derived from the remediation plan already. And after checking further, getSuccessfulClusters will not return clusters where backup or precaching is in progress which was what was causing my issues with the empty lists before

PR #353, opened 2022-10-05
OCPBUGS-1903: Do not copy labels from the original configmap

Missxiaoguo:


donpenney (commented):
/lgtm

PR #356, opened 2022-10-07
CNF-6462: code-gen: add scripts, update code and upload generated clientset, listers and informers code

pixelsoccupied:
Clientset, Listers, Informers generated code for easier development of TALM functional tests


api/v1alpha1/groupversion_info.go
```diff
@@ -28,6 +28,9 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "ran.openshift.io", Version: "v1alpha1"}
 
+	// SchemeGroupVersion is group version used to register these objects. This is expected by code-gen
+	SchemeGroupVersion = schema.GroupVersion{Group: "ran.openshift.io", Version: "v1"}
```
jc-rh, line 32: is it intentional to use v1 rather than v1alpha1?

pixelsoccupied: Yep should be v1alpha1!

hack/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+bash ${CODEGEN_PKG}/generate-groups.sh "all" \
+    "github.com/openshift-kni/cluster-group-upgrades-operator/pkg/generated" \
+    "github.com/openshift-kni/cluster-group-upgrades-operator/api" \
+    clustergroupupgradesoperator:v1alpha1 \
+    --output-base "./pkg" \
+    --go-header-file hack/boilerplate.go.txt
+
+cp -rp "./pkg/github.com/openshift-kni/cluster-group-upgrades-operator/pkg/" "./pkg/"
+
+sed -i '' "s|api/clustergroupupgradesoperator/v1alpha1|api/v1alpha1|g" $(find pkg/generated -type f)
```
donpenney, line 27: The '' is causing your regex to be seen as a file argument, causing the following error:
sed: can't read s|api/clustergroupupgradesoperator/v1alpha1|api/v1alpha1|g: No such file or directory

In turn, this causes update-codegen.sh to fail with exit code 2.

hack/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# more info: https://cloud.redhat.com/blog/kubernetes-deep-dive-code-generation-customresources
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+mkdir -p ./api/clustergroupupgradesoperator
+cd ./api/clustergroupupgradesoperator
+ln -s ../v1alpha1 ./v1alpha1
```
donpenney, line 11: Is api/clustergroupupgradesoperator intended to be a working dir that doesn't already exist and shouldn't be committed? In the case of failure, the script is exiting before this is cleaned up at the end of the script, and this ln command, for example, fails.

hack/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+    --output-base "./pkg" \
+    --go-header-file hack/boilerplate.go.txt
+
+cp -rp "./pkg/github.com/openshift-kni/cluster-group-upgrades-operator/pkg/" "./pkg/"
+
+sed -i '' "s|api/clustergroupupgradesoperator/v1alpha1|api/v1alpha1|g" $(find pkg/generated -type f)
+
+# clean up
+rm -rf ./pkg/github.com/
+rm -rf api/clustergroupupgradesoperator
```
donpenney, line 31: Consider a cleanup function in a trap on EXIT. Because of the errexit at the beginning of the script, any failure (such as the sed command above) means this cleanup doesn't occur.

donpenney (changes requested):
Is the update-codegen.sh expected to be called manually by developers, with updated generated code committed? If so, it would probably be good to add it to the Makefile, probably to the "generate" target as a dependency, to ensure it doesn't get missed and files go stale. I don't see any Makefile changes here, unless there's some other hook that triggers this already?

hack/update-codegen.sh
pixelsoccupied: Added trap for cleanup

hack/update-codegen.sh
pixelsoccupied: Yeah it's intended to be like this i.e used just to keep code-gen happy and not committing. But added trap for cleanup.

hack/update-codegen.sh
pixelsoccupied: Sorry about that. Ya needed to add '' for macOS cause [apparently](https://stackoverflow.com/questions/16745988/sed-command-with-i-option-in-place-editing-works-fine-on-ubuntu-but-not-mac) sed works differently there. Added better support. Can you give it shot again?

hack/update-codegen.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# TIP: append '-v 10' with the script call below for logs
+bash ${CODEGEN_PKG}/generate-groups.sh "all" \
+    "github.com/openshift-kni/cluster-group-upgrades-operator/pkg/generated" \
+    "github.com/openshift-kni/cluster-group-upgrades-operator/api" \
+    clustergroupupgradesoperator:v1alpha1 \
+    --output-base "./pkg" \
+    --go-header-file hack/boilerplate.go.txt
+
+cp -rp "./pkg/github.com/openshift-kni/cluster-group-upgrades-operator/pkg/" "./pkg/"
```
donpenney, line 35: I'm left with a pkg/pkg untracked dir after running this script. This cp command appears to just be copying the generated pkg directory over to the pkg dir.

If I change this to use "./" instead, I think it does what you expect. But now I'm seeing some uncommitted changes in two generated files:
        modified:   pkg/generated/clientset/versioned/fake/register.go
        modified:   pkg/generated/clientset/versioned/scheme/register.go

Adding a call to the update-codegen.sh script to the "generate" target of the Makefile would help catch uncommitted changes in these generated files, as we have CI checks to ensure any such generated changes are committed.

donpenney (changes requested):

hack/update-codegen.sh
pixelsoccupied: Nice! Ok added support for Makefile.

pixelsoccupied: Added the CI config. 

`register.go` is coming out differently on macos vs when I run it on rhel. Interesting. The latest pushed code now should be correctly spaced.

Thank you for the `rsync` code :D

donpenney (approved):
/lgtm

PR #363, opened 2022-10-19
OCPBUGS-2595: Fix issue where failed precaching/backup clusters were not removed from further remediation consideration

fontivan:
* After precaching, any clusters that failed precaching will be removed from further remediation (backup and upgrade)
* After backup, any clusters that failed backup will be removed from further remediation (upgrade)

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -266,8 +266,10 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				return
 			}
 		}
-
 	}
+
+	// Update the clusters list based on the precaching results
+	clusters = r.filterFailedPrecachingClusters(ctx, clusterGroupUpgrade, clusters)
 
 	suceededCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.Succeeded))
 
@@ -366,6 +368,9 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 				}
 			}
 
+			// Update the clusters list based on the backup results
+			clusters = r.filterFailedBackupClusters(ctx, clusterGroupUpgrade, clusters)
+
 			backupCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.BackupSuceeded))
 			if clusterGroupUpgrade.Status.Backup == nil || (backupCondition != nil && backupCondition.Status == metav1.ConditionTrue) {
 
@@ -380,6 +385,11 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					}
 					if allManagedPoliciesExist {
 						r.buildRemediationPlan(clusterGroupUpgrade, clusters, managedPoliciesInfo.presentPolicies)
+
+						// Rebuild clusters list since remediation plan may have changed
+						clusters = r.getClustersListFromRemediationPlan(clusterGroupUpgrade)
+						clusters = r.filterFailedPrecachingClusters(ctx, clusterGroupUpgrade, clusters)
+						clusters = r.filterFailedBackupClusters(ctx, clusterGroupUpgrade, clusters)
 					}
 				}
 				// Take actions before starting upgrade.
@@ -406,6 +416,19 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					)
 					nextReconcile = requeueImmediately()
 				} else {
+
+					// Check that we still have clusters to remediate since errors may have occured above
+					if len(clusters) == 0 {
+						utils.SetStatusCondition(
+							&clusterGroupUpgrade.Status.Conditions,
+							utils.ConditionTypes.Progressing,
+							utils.ConditionReasons.NotStarted,
+							metav1.ConditionFalse,
+							"Unable to begin upgrade due to precaching/backup errors",
+						)
+						nextReconcile = requeueWithLongInterval()
+					}
+
 					// Start the upgrade.
 					utils.SetStatusCondition(
 						&clusterGroupUpgrade.Status.Conditions,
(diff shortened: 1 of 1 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -406,6 +416,20 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					)
 					nextReconcile = requeueImmediately()
 				} else {
+
+					// Check that we still have clusters to remediate since errors may have occured above
+					if len(clusters) == 0 {
```
sabbir-47, line 421: We don't need this progressing condition message, if backup failed for all the cluster as an example, backupCondition.Status will be metav1.ConditionFalse, preventing us from coming to this block

controllers/clustergroupupgrade_controller.go
```diff
@@ -380,6 +385,11 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					}
 					if allManagedPoliciesExist {
 						r.buildRemediationPlan(clusterGroupUpgrade, clusters, managedPoliciesInfo.presentPolicies)
+
+						// Rebuild clusters list since remediation plan may have changed
+						clusters = r.getClustersListFromRemediationPlan(clusterGroupUpgrade)
+						clusters = r.filterFailedPrecachingClusters(ctx, clusterGroupUpgrade, clusters)
+						clusters = r.filterFailedBackupClusters(ctx, clusterGroupUpgrade, clusters)
```
sabbir-47, line 392: Rather than calling all these functions, after building the remediation plan, can't we check if it is empty or not? If a remediation plan is empty, we can set a condition message, we don't have to look at the length of the cluster

controllers/clustergroupupgrade_controller.go
```diff
@@ -380,6 +385,11 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					}
 					if allManagedPoliciesExist {
 						r.buildRemediationPlan(clusterGroupUpgrade, clusters, managedPoliciesInfo.presentPolicies)
```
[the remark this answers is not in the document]
fontivan: It does not, thats why I am using the filtering functions just after the build

controllers/clustergroupupgrade_controller.go
fontivan: The remediation plan is built based on the various selectors in the CGU, but it would not account for failed precaching/backup clusters, so I filtered them here

controllers/clustergroupupgrade_controller.go
fontivan: Actually yeah we can only get here if the backup was at least partially successful, so I don't need this check at all

controllers/clustergroupupgrade_controller.go
jc-rh: I made the same comment on Sabbir's previous PR. This whole block is a duplicate of some code earlier in the function. Now we are forced to duplicate the filtering calls as well. It would be cleaner if we simply requeue after backup is done.

fontivan: Actually reading this again, I don't think I need to re-filter or that we need to recheck the policies
We should just be able to rebuild the remediation plan and continue since the clusters will have already been filtered and the policies would have already been checked

controllers/clustergroupupgrade_controller.go
```diff
@@ -252,6 +252,13 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}
 
+	if len(clusters) == 0 && len(clusterGroupUpgrade.Status.RemediationPlan) != 0 {
```
jc-rh, line 255: Why is this check needed here? I don't see filtering prior to this.

controllers/clustergroupupgrade_controller.go
```diff
@@ -342,6 +351,14 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 		} else {
 			// There are no blocking CRs, continue with the upgrade process.
 			// create backup
+
```
jc-rh, line 354: Does an empty cluster list break backup or other operations below? Also, how does waiting help? We get here only if precaching is no longer active.

controllers/clustergroupupgrade_controller.go
jc-rh: But this is before precaching. The expression will never be true?

controllers/clustergroupupgrade_controller.go
jc-rh: I think that's ok, as long as the status clearly shows what has happened. I was wondering whether we should have another variant for the succeeded condition that indicates clusters being skipped due to backup/precaching failures. WDYT @sabbir-47 @fontivan ?

controllers/clustergroupupgrade_controller.go
fontivan: I've removed it

controllers/clustergroupupgrade_controller.go
fontivan: If we hit this condition it means that precaching failed for all clusters so we definitely don't want it to then claim that backup was successful when it was never run

sabbir-47: Why should never send an empty list to precaching/backup if clusterList actually is not empty.

controllers/clustergroupupgrade_controller.go
sabbir-47: if line 153 returns empty list of clusters, does line 187 returns true for allManagedPoliciesExist? if it returns true, buildremediation plan will have empty remediation plan, and clusters will be empty in line 213.

controllers/clustergroupupgrade_controller.go
sabbir-47: @jc-rh if empty list is sent to backup/precaching the first time, they will never be executed, although it sends the clusterList in the second reconcile.

jc-rh: Since waiting longer won't change anything, we should just add the final succeeded condition as false with a reason like "no cluster available "

sabbir-47: I would rather add that message in the progressing condition.

controllers/clustergroupupgrade_controller.go
```diff
@@ -366,29 +383,39 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					metav1.ConditionFalse,
+					"No clusters available for remediation (Backup failed)",
+				)
+				// Requeue is required here since we need to come back and check the succeeded condition for final cleanup
+				nextReconcile = requeueImmediately()
+				return
+			}
+
 			backupCondition := meta.FindStatusCondition(clusterGroupUpgrade.Status.Conditions, string(utils.ConditionTypes.BackupSuceeded))
 			if clusterGroupUpgrade.Status.Backup == nil || (backupCondition != nil && backupCondition.Status == metav1.ConditionTrue) {
```
jc-rh, line 406: We said we didn't need this right? Plus, there is no else block.

jc-rh:
Need similar handling for the new succeeded cases we added
```r.Recorder.Event(clusterGroupUpgrade, corev1.EventTypeWarning, "UpgradeTimedOut", "The ClusterGroupUpgrade CR policies are taking too long to complete")
r.Log.Info("CGU has timed out")```

**Comment by fontivan** [comment]
/hold
must retest latest changes

File: controllers/clustergroupupgrade_controller.go [hunk 7]
```diff
@@ -287,10 +303,10 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			}
 		} else {
 			// Upgrade has failed
-			r.Recorder.Event(clusterGroupUpgrade, corev1.EventTypeWarning, "UpgradeTimedOut", "The ClusterGroupUpgrade CR policies are taking too long to complete")
-			r.Log.Info("CGU has timed out")
+			r.Recorder.Event(clusterGroupUpgrade, corev1.EventTypeWarning, suceededCondition.Reason, suceededCondition.Message)
+			r.Log.Info("CGU has failed")
```
jc-rh, line 307: Why not just print the reason. Not sure if event gets logged as well.

jc-rh (approved):

PR #364, opened 2022-10-20
Avoid panics in edge cases

jc-rh:
These cases happened when a really old TALM image got deployed accidentally. Panics from various places due to the mismatch between API and binary. CGUs got left behind in weird states which caused more panics even with the right version of TALM. It's difficult to cleanup them when TALM doesn't stay up.


donpenney (commented):
/lgtm

nishant-parekh (approved):

PR #368, opened 2022-10-20
OCPBUGS-2072: CGU-Hangs-When-A-ManagedCluster-is-Unavailable

sabbir-47:
- Not reachable cluster case handled for backup
- Introduced timeout for backup that prevent CGU from hanging
- Remaining MCA are deleted for non reachable clusters

/cc @jc-rh @fontivan


jc-rh:
Obsoleted by #377 and #383

PR #371, opened 2022-10-28
OCPBUGS-2657: CGU Insufficient disk space to create backup

sabbir-47:
- SafetyNet unit declared
- Unit test added

/cc @imiller0 @donpenney @jc-rh

### DIFF
--- a/recovery/cmd/backup_size_test.go
+++ b/recovery/cmd/backup_size_test.go
@@ -1,0 +1,121 @@
+package cmd_test
+
+import (
+	"testing"
+
+	"github.com/openshift-kni/cluster-group-upgrades-operator/recovery/cmd"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCompare(t *testing.T) {
+
+	testcases := []struct {
+		estimated, freeDisk float64
+		expected            bool
+		name                string
+	}{
+		{
+			estimated: 80.00 * 1024 * 1024 * 1024,
+			freeDisk:  100.00 * 1024 * 1024 * 1024,
+			expected:  true,
+			name:      "higher disk space in same unit",
+		},
+		{
+			estimated: 800.00 * 1024 * 1024 * 1024,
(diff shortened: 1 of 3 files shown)


recovery/cmd/backup_size_test.go
```diff
@@ -0,0 +1,121 @@
...
+			freeDisk:  50.00 * 1024 * 1024 * 1024,
+			expected:  false,
+			name:      "lower disk space in same unit",
+		},
+		{
+			estimated: 80.00 * 1024 * 1024 * 1024,
+			freeDisk:  100.00 * 1024 * 1024,
+			expected:  false,
+			name:      "lower disk space in different unit",
+		},
```
donpenney, line 46: Probably be good to add a case here where difference is units is at least two levels. It looks like these are all either same units or off by one (ie. G vs M), but the case that came up was terabytes of free space vs megabytes of required space. I think the original code was ok with the off-by-one (aside from the safety net issue), but had an issue with the off-by-two

recovery/cmd/backup_size_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func TestCompare(t *testing.T) {
+
+	testcases := []struct {
+		estimated, freeDisk float64
+		expected            bool
+		name                string
+	}{
+		{
+			estimated: 80.00 * 1024 * 1024 * 1024,
```
donpenney, line 18: nit: Can you add comments to each of these to explicitly call out the units/sizes for each test?

recovery/cmd/backup_size_test.go
sabbir-47: Added by this commit 0a94f2debff8eb3ed881ec1eab12431501f50218 as suggested

recovery/cmd/backup_size_test.go
sabbir-47: Added by this commit https://github.com/openshift-kni/cluster-group-upgrades-operator/commit/0a94f2debff8eb3ed881ec1eab12431501f50218 as suggested

PR #377, opened 2022-11-02
Handle cluster errors during backup

jc-rh:


controllers/clustergroupupgrade_controller.go
```diff
@@ -335,6 +336,9 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			return
 		}
 
+		if clusterGroupUpgrade.Status.Status.StartedAt.IsZero() {
+			clusterGroupUpgrade.Status.Status.StartedAt = metav1.Now()
+		}
```
fontivan, line 341: won't this cause an issue when the cgu starts later, since this will be used to check for cgu timeout?
I would instead expect the backup started at field you added to be initialized somewhere

jc-rh: It's initialized here:
```
	if clusterGroupUpgrade.Spec.Backup {
		// Backup is required
		if clusterGroupUpgrade.Status.Backup == nil {
			clusterGroupUpgrade.Status.Backup = &ranv1alpha1.BackupStatus{
				Status:    make(map[string]string),
				Clusters:  []string{},
				StartedAt: metav1.Now(),
			}
		}
```

jc-rh: Not sure what you mean by later. CGU starts before backup starts.

PR #379, opened 2022-11-04
OCPBUGS-3286: Skip clusters not on subscription policies properly

jc-rh:


Missxiaoguo (approved):
Looks good

controllers/monitoring.go
```diff
@@ -191,12 +191,10 @@ func (r *ClusterGroupUpgradeReconciler) processMonitoredObjects(
...
 		}
 		managedPolicyName := clusterGroupUpgrade.Status.ManagedPoliciesForUpgrade[*clusterProgress.PolicyIndex].Name
-
-		// If there is no monitored object saved for the current managed policy, return.
 		_, ok := clusterGroupUpgrade.Status.ManagedPoliciesContent[managedPolicyName]
 		if !ok {
-			r.Log.Info("[approveInstallPlan] No content for policy", "managedPolicyName", managedPolicyName)
-			return false, nil
+			r.Log.Info("[processMonitoredObjects] No monitored object found for policy", "managedPolicyName", managedPolicyName)
+			continue
```
imiller0, line 197: Are there failure cases where ManagedPoliciesContent does not get populated that we may miss with this fix?
It would also be helpful to have a comment here describing why we allow this case (since it looks like an error).

jc-rh: No. How about this log message instead:
```
r.Log.Info("[processMonitoredObjects] Current policy for cluster " + clusterName + " does not contain monitored object.")
```

PR #394, opened 2022-11-29
OCPBUGS-4197: Handle error responses from upgrade graph server

jc-rh:


pixelsoccupied:
Tried on a live cluster too, error is now handled correctly. 

Just for my knowledge (first time reviewing this project's PR) -- we are not expecting any unit test for this fix or `getImageForVersionFromUpdateGraph` in general?

jc-rh:
> Tried on a live cluster too, error is now handled correctly.
> 
> Just for my knowledge (first time reviewing this project's PR) -- we are not expecting any unit test for this fix or `getImageForVersionFromUpdateGraph` in general?

In general, yes. For this function though, it would require mocking an http response. I will see if I can find some examples and add it later.

PR #399, opened 2022-12-07
OCPBUGS-3779: support only image or version in ClusterVersion when doing ocp precache

pixelsoccupied:
- support clusterVersion CR with only Image or only version or both. 
- reduce unnecessary calls by making:
         - single external only after CR validation
         - no calls if user provides the Spec.DesiredUpdate.image
 - unit tests 
 
/cc @jc-rh @fontivan

### DIFF
--- a/controllers/validation_test.go
+++ b/controllers/validation_test.go
@@ -1,0 +1,242 @@
+package controllers
+
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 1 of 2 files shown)


controllers/validation_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"k8s.io/client-go/tools/record"
+	"net/http"
+	"net/http/httptest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+func TestClusterGroupUpgradeReconciler_extractOpenshiftImagePlatformFromPolicies(t *testing.T) {
+
+	policy := `---
```
fontivan, line 19: these policies should probably all be const no?

pixelsoccupied: Most actually. One of them needs to be variable since it has to make fake http call. But changed the rest to const (where there is no https calls)

jc-rh (approved):

PR #404, opened 2023-01-11
Check node space available before running precache

sudomakeinstall2:
checks that node has more than 15 GB before it reaches GChighThreshold
to avoid triggering garbage collection.
Only do this when the cgu contains an OCP upgrade.
15 is a rough estimate. 10 GB for OCP upgrade and 5 GB for operators, etc


controllers/templates/precache-templates.go
```diff
@@ -136,6 +136,12 @@ spec:
               env:
               - name: config_volume_path
                 value: /tmp/precache/config
+              - name: SPACE_REQUIRED
```
jc-rh, line 139: @vitus133 Do you think we should add these to the existing config volume instead of more env vars?

pre-cache/check_space
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+
+APISERVER=https://kubernetes.default.svc
+SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
+NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
+TOKEN=$(cat ${SERVICEACCOUNT}/token)
+CACERT=${SERVICEACCOUNT}/ca.crt
+curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X GET ${APISERVER}/api/v1/nodes/${NODE_NAME}/proxy/configz > /tmp/kubeletconfig
```
jc-rh, line 9: @vitus133 Does the node name really have to be passed in? I imagine it must be on the node itself somewhere?

controllers/utils/constants.go
```diff
@@ -137,3 +137,7 @@ const (
 	PlcHubTmplPrinfInNameErr = "printf variable is not supported in the template function Name field"
 	PlcHubTmplPrinfInNsErr   = "printf variable is not supported in the template function Namespace field"
 )
+
+// SpaceRequiredForPrecache is an env variable for precaching job that indicates the amount of space required
+// 							for precahing job
+const SpaceRequiredForPrecache = "15" // Gigabytes
```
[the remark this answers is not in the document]
jc-rh: @vitus133 Should we use a different number if it's operator upgrade only? Or skip this step altogether?

controllers/templates/precache-templates.go
vitus133: I think using downwards API with env vars is a clean way of passing info to pods.  It is also easy to template, so in the future, when we determine the precache spec (and the payload size) on the hub, this info will be easy to pass forward.

pre-cache/check_space
vitus133: I think only the hostname is on the node. Not necessarily equals to the node name. Using downward API for passing the node name is one  of the examples given in the k8s documentation, so I assume it is also the recommended way
https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

controllers/templates/precache-templates.go
jc-rh: Ok, so we could get rid of the config map and pass everything this way? That would save us an MCV and an MCA?

controllers/utils/constants.go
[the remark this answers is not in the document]
vitus133: I think there is no typical number, because it depends on the user's use case.
I think we need to template this number sooner the better. It means we should accelerate extracting the precaching spec on the hub

vitus133:
I left a couple of comments above. We can merge it now, wait until we implement the precaching spec creation on the hub, or, add some functionality for obtaining the payload size on the spoke.
What do you guys think?

controllers/templates/precache-templates.go
vitus133: Yes, it's possible. The CM is somewhat more convenient for creating several independent files, where potentially we can have structured text

controllers/utils/constants.go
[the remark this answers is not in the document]
jc-rh: @imiller0 @browsell As we discussed, let's keep it simple and check the disk space against a static size for OCP upgrade. For operator upgrade, there isn't a good default to use. I imagine it would be a way smaller number. Is it small enough to be ignored?

pre-cache/precache.sh
```diff
@@ -5,6 +5,8 @@ set -e
 rm -rf /host/tmp/precache
 cp -a /opt/precache /host/tmp/
 cp -rf /etc/config /host/tmp/precache/config
+/opt/precache/check_space
+[ $? -ne 0 ] && echo "not enough space for precaching" && exit 17
```
jc-rh, line 9: There doesn't seem to be a way to pass the return code back to the job. The error will be reported as a generic error in the CGU. @vitus133 @imiller0 @browsell Any objection or alternative? Splitting this into its own job seems to be a bit too much for just a better error message?

imiller0: Are there meaningful log messages that can be looked at to determine the failure reason?

sudomakeinstall2: yes but you have to look at precache pod logs

pre-cache/check_space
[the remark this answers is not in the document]
sudomakeinstall2: fixed

controllers/utils/constants.go
```diff
@@ -137,3 +137,7 @@ const (
 	PlcHubTmplPrinfInNameErr = "printf variable is not supported in the template function Name field"
 	PlcHubTmplPrinfInNsErr   = "printf variable is not supported in the template function Namespace field"
 )
+
+// SpaceRequiredForPrecache is an env variable for precaching job that indicates the amount of space required
+// for precahing job. This is an rough estimate. 10 GB for OCP images and 5 GB for operator images.
```
jc-rh, line 142:
```suggestion
// for precaching job. This is a rough estimate. 10 GB for OCP images and 5 GB as a buffer for operator images and other disk allocations that can happen between the job and the actual upgrade kicked off by the CGU.
```

sudomakeinstall2:
/test integration
--


<br class="Apple-interchange-newline">

controllers/utils/constants.go
jc-rh: This is not incorporated?

sudomakeinstall2: I think it got lost when I merged the master into it :))

jc-rh (approved):

PR #405, opened 2023-01-11
Execute afterCompletionAction on a per cluster basis

jc-rh:
This also fixes the already compliant cluster cases for both labelling and cluster status update.


fontivan (approved):
/approve

fontivan (approved):
/approve

PR #406, opened 2023-01-12
Stop recreating backup status map

jc-rh:
We lose clusters when the cluster list changes in edge scenarios. This change is similar to #390.


controllers/backup.go
```diff
@@ -106,13 +104,12 @@ func (r *ClusterGroupUpgradeReconciler) triggerBackup(ctx context.Context,
 			nextState = BackupStateTimeout
 		}
 
-		clusterStates[cluster] = nextState
+		clusterGroupUpgrade.Status.Precaching.Status[cluster] = nextState
```
sabbir-47, line 107: Why are we populating precaching status map and not backup status map? @jc-rh

jc-rh:
Wish you were here last week. No one noticed this copy/paste error. It's
fixed now.

PR #409, opened 2023-01-14
OCPBUGS-5797: Fix a copy/paste error

jc-rh:


jc-rh:
Merging in case it can still make into 4.12.0

PR #420, opened 2023-02-01
OCPBUGS-2812: Do not skip compliant policies containing status fields

jc-rh:


Missxiaoguo:
With this change, the compliant subscription policy will not be skipped at the beginning. Wondering if you tested in real? Can the subscription policy become NonCompliant before TALM gets into it?

jc-rh:
> The code looks good to me. With this change, the compliant subscription policy will not be skipped at the beginning. Wondering if you tested in real? Can the subscription policy become NonCompliant before TALM gets into it?

No. I did not. This fix is as good as the annotation workaround. The race condition is tracked by OCPBUGS-2813. In the scale lab, I've seen the catalog source alone taking more than 12 minutes to become ready again: https://redhat-internal.slack.com/archives/C03QX2FG4JW/p1674145146463829

PR #428, opened 2023-02-06
OCPBUGS-4821: Make precache graph call secure by default

sudomakeinstall2:


sudomakeinstall2:
@vitus133 does this work correctly? or should we pull the graph from somewhere else?

pixelsoccupied:
IMO we should leave Insecure for the sake of backwards compatibility and ease of use when environment is locked up and allow it toggle with an env var `TALM_INSECURE`. And additionally support secure call on top this. 

when you don't allow insecure...that's when the user would need to provide a cert. So like your go code need to be able read the crt uploaded to by user maybe in ConfigMap and then you read it with `x509.NewCertPool()` and so on. I think Vitaly mentioned in the comments. 

here's an example of secure call code https://stackoverflow.com/a/38825553/9070865 . 

I think MCO and other operators are also doing something similar...would refer to those codebase for some tips too :)

pixelsoccupied (commented):
This is nice! 


a small nit + it would be great to update the manifests to reflect this ENV variable. I think it would should show up in the deployment manifest like [others](https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/138e261337fd52f4674efb6ebbdd4b4587cb363f/bundle/manifests/cluster-group-upgrades-operator.clusterserviceversion.yaml#L359)? IMO hardcoding this is fine (as opposed to putting something in makefile?) as we want to be secure default but up to you. But with this devs/users are aware about how to flip the switch

pixelsoccupied:
Also imo beyond the scope of this PR but just so I don't forget...here's the example code i was talking about earlier (re: custom cert) from CVO (they are calling graph url also) 

[getTransport](https://github.com/openshift/cluster-version-operator/blob/4f130ae8097238ebec5186996346a47c70299eb0/pkg/cvo/egress.go#L28-L52) which calls [getTLSConfig](https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/egress.go#L73-L97) where a cm with cert is read (and falls back to what you now in L73). Bonus it looks like they even allow users to proxy. 

I think it's a worth while have discussion to up the configurability and/or security.

jc-rh (approved):

PR #434, opened 2023-02-08
OCPBUGS-6944: recovery: Update container checks in upgrade-recovery.sh

donpenney:
The upgrade-recovery.sh tool uses "crictl ps" to check container status, but the format of the ps output has been updated. Rather than screen-scraping for a specific column, this update changes the script to use "crictl ps -o json", using "jq" to pull the desired data from the output.

/cc @jc-rh @Missxiaoguo @sabbir-47


Missxiaoguo (commented):
/lgtm

PR #438, opened 2023-02-09
OCPBUGS-7217: Keep conditions in order of the status transition time

sudomakeinstall2:
When setting condition, if the condition status is changed it should be the last condition in the list, so that printercolumn works as expected


jc-rh:
Add a kuttl test that does the scenario that exposed the bug?

sudomakeinstall2:
> Add a kuttl test that does the scenario that exposed the bug?

I think adding a kuttl test is overkill. But I added unit test for that function.

jc-rh (approved):

PR #479, opened 2023-02-21
OCPBUGS-9395: backup documentation updated for custom partition

sabbir-47:
/cc @jc-rh @alosadagrande

### DIFF
--- a/docs/backup-recovery/README.md
+++ b/docs/backup-recovery/README.md
@@ -14,7 +14,7 @@ There are two methods related to partition creation is described here:
 
 The recovery partition can be created at install time by defining a node level property called `diskPartition` in the SiteConfig CR. The configuration for the partition must be defined by `start` parameter which must indicate, in mebibytes, the end of the `/sysroot` RHCOS partition and the start for the newly requested recovery partition.
 
-Additionally, a spare disk can be used for recovery partition as well and can be configured at provisioning time or at day 2 operation. A recovery partition of 50GB with mountpoint at `/var/recovery` can be created by following the below example. Notice that the deviceName can be referenced by its path, label or id as well, e.g. `/dev/disk/by-id/scsi-355cd2e404b7c23f3`
+Additionally, a spare disk can be used for recovery partition as well and can be configured at provisioning time or at day 2 operation. A recovery partition of 50GB with mountpoint at `/var/recovery` can be created by following the below example. Notice that the device under diskPartition must be referenced by a persistent path, such as those under /dev/disk/by-path or /dev/disk/by-id (where the by-path name would be recommended, as it is not hardware-specific and would not change in the case of disk replacement)
 
 ```yaml
 nodes:
@@ -27,7 +27,7 @@ nodes:
         ........
         #Disk /dev/sda: 893.3 GiB, 959119884288 bytes, 1873281024 sectors
         diskPartition:
-          - device: /dev/sda
+          - device: /dev/disk/by-path/pci-0000:18:00.0-scsi-0:2:1:0
             partitions:
               - mount_point: /var/recovery
                 size: 51200


docs/backup-recovery/README.md
```diff
@@ -14,7 +14,7 @@ There are two methods related to partition creation is described here:
 
 The recovery partition can be created at install time by defining a node level property called `diskPartition` in the SiteConfig CR. The configuration for the partition must be defined by `start` parameter which must indicate, in mebibytes, the end of the `/sysroot` RHCOS partition and the start for the newly requested recovery partition.
 
-Additionally, a spare disk can be used for recovery partition as well and can be configured at provisioning time or at day 2 operation. A recovery partition of 50GB with mountpoint at `/var/recovery` can be created by following the below example. Notice that the deviceName can be referenced by its path, label or id as well, e.g. `/dev/disk/by-id/scsi-355cd2e404b7c23f3`
+Additionally, a spare disk can be used for recovery partition as well and can be configured at provisioning time or at day 2 operation. A recovery partition of 50GB with mountpoint at `/var/recovery` can be created by following the below example. Notice that the deviceName must be referenced by its id, e.g. `/dev/disk/by-id/scsi-355cd2e404b7c23f3`
```
donpenney, line 17: I would assume any of the persistent names should be ok, not just id. Since id is specific to a particular piece of hardware, using the by-path name (/dev/disk/by-path) is both generic and persistent. ie. The same value can be used for different nodes with the same configuration. So replacing a disk doesn't require a change in the site-config, and the same site-config subsection can be cloned across all nodes with the same disk configuration

sabbir-47: The texts were modified by following the bug. https://issues.redhat.com/browse/OCPBUGSM-46835 
I can add both id and by path.

donpenney: I think the intent of the bug is primarily to avoid using the named device like sdX (I'm not sure how safe nvme names would be), as you can easily have devices named differently over a reboot. I fully agree with avoiding using those device names. But my personal preference is to use by-path rather than by-id or by-uuid as it is both persistent and generic, and would therefore be much better in lab config templates. And would not be impacted by hardware replacements (ie. replacing a bad disk would not require a change to the site-config)

josclark42: I agree with @donpenney 's suggestion to  change the doc to specify one of the persistent device names. Using by-path makes sense to me.

alosadagrande: I do not have a strong opinion about which one is better either by-path or by-id looks good to me. Avoiding device is needed.

alosadagrande: https://docs.openshift.com/container-platform/4.12/rest_api/provisioning_apis/baremetalhost-metal3-io-v1alpha1.html#spec-rootdevicehints

alosadagrande (commented):
I do not have a strong opinion about which one is better either by-path or by-id looks good to me. Avoiding device name is needed.

docs/backup-recovery/README.md
sabbir-47: @alosadagrande if we don't set deviceName, is hctl enough under rootDeviceHints?

sabbir-47: It seems WWID is reliable and unique as explained here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/storage_administration_guide/persistent_naming#persistent_naming-wwid

And spec.rootDeviceHInts accepts wwn value. Need to be confirmed with someone who has expertise.

sabbir-47: Changed to "by-path"

sabbir-47: @donpenney changed as suggested. may be a lgtm?

jc-rh (approved):

PR #481, opened 2023-02-23
OCPBUGS-7540: Wait for `soakSeconds` after policy becomes compliant before moving on from policy

sudomakeinstall2:
If a policy has the `soakSeconds` annotation, the controller will wait for the specified
number of seconds after the first time it observes that the cluster is compliant with
that policy before moving on from that policy. The controller will set
`status.status.currentBatchRemediationProgress.clusterName.firstCompliantAt` the first time
it observes that the cluster is compliant with the policy. In the following reconciliations,
if the requested number of seconds has passed since firstCompliantAt, and the cluster remains
compliant with the policy, the controller will move on from that policy.


tests/kuttl/tests/soaking/03-complete-batch2.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+
+commands:
+  # Patch the inform policies to reflect the compliance status Compliant for spoke1.
```
jc-rh, line 5: Update the comment and the file name to reflect what it really does in this test

jc-rh (approved):

PR #494, opened 2023-03-06
OCPBUGS-8006: More frequent readiness check for new cluster only

jc-rh:
Also removed repetitive logs.

### DIFF
--- a/controllers/managedclusterForCGU_controller.go
+++ b/controllers/managedclusterForCGU_controller.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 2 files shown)


controllers/managedclusterForCGU_controller.go
```diff
@@ -134,13 +131,19 @@ func (r *ManagedClusterForCguReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 	} else {
 		// availableCondition is false or unknown, cluster is not ready
-		r.Log.Info("cluster is not ready", "RequeueAfter:", clusterStatusCheckRetryDelay)
```
nishant-parekh, line 137: do you want to add this log to the new common function, probably helps with debugging?

nishant-parekh (approved):
just a small suggestion. but looks good otherwise

controllers/managedclusterForCGU_controller.go
jc-rh: Potentially. Normally this log is just flooding our log file, especially with this change. There seem to be only 2 levels with our logger: info and error.

PR #555, opened 2023-05-10
Validate dependencies order

sudomakeinstall2:
For each managed policies that should be remediated, check its dependencies and verify non of its dependencies comes later in the managed policies list.


controllers/validation.go
```diff
@@ -121,3 +122,41 @@ func (r *ClusterGroupUpgradeReconciler) validateOpenshiftUpgradeVersion(
...
+		specObject := managedPolicy.Object["spec"].(map[string]interface{})
+		dependencies := specObject["dependencies"]
+		if dependencies == nil {
+			continue
+		}
+		dependenciesArr := dependencies.([]interface{})
+
+		for _, d := range dependenciesArr {
+			name := d.(map[string]interface{})["name"]
+			dependecyIndex, err := indexOf(name.(string), clusterGroupUpgrade.Spec.ManagedPolicies)
```
jc-rh, line 148: We should verify the namespace too.

controllers/validation.go
```diff
@@ -121,3 +122,41 @@ func (r *ClusterGroupUpgradeReconciler) validateOpenshiftUpgradeVersion(
...
+func indexOf(element string, data []string) (int, error) {
+	for k, v := range data {
+		if element == v {
+			return k, nil
+		}
+	}
+	return -1, errors.New("element not found in data")
+}
+
+func (r *ClusterGroupUpgradeReconciler) validatePoliciesDependenciesOrder(
```
irinamihai, line 135: Why don't we automatically order managed policies to satisfy dependencies?

sudomakeinstall2: @jc-rh what do you think about this?
It makes sense to me.

jc-rh: What if there is a circular dependency? I would rather flag it and let user decide what to fix. We can discuss what is the best way to let user specify a policy set with built-in dependency. It seems to be cleaner to have a different spec field rather than switching the order in the existing list.

irinamihai: I did consider circular dependencies and thought we could have a warning for that, to help the user. I think it's fine to leave it as is for now and keep this in mind for the future.

jc-rh: Sure, feel free to start capturing potential use cases in https://github.com/jc-rh/kni-enhancements/blob/main/enhancements/talm-acm-alignment-policy-dependency-support.md
The scope of this change is to make sure TALM still functional with policies with dependency.

controllers/validation.go
```diff
@@ -121,3 +122,46 @@ func (r *ClusterGroupUpgradeReconciler) validateOpenshiftUpgradeVersion(
...
+			dependecyIndex, err := indexOf(
+				ranv1alpha1.ManagedPolicyForUpgrade{Name: name.(string), Namespace: namespace.(string)},
+				clusterGroupUpgrade.Status.ManagedPoliciesForUpgrade)
+			if err == nil && dependecyIndex > managedPolicyIndex {
+				utils.SetStatusCondition(
+					&clusterGroupUpgrade.Status.Conditions,
+					utils.ConditionTypes.Validated,
+					utils.ConditionReasons.InvalidDependencyOrder,
+					metav1.ConditionFalse,
+					fmt.Sprintf("Invalid Dependecy Order, Managed Policy %s is dependent on %s, but the dependency comes earlier than the managed policy in ManagedPolicies list", managedPolicy.GetName(), name),
```
jc-rh, line 160: The first part is redundant since it's already in in the reason? Also, not sure how to improve, but the sentence is a bit hard to read.

jc-rh: How about:

```suggestion
					fmt.Sprintf("Managed Policy %s is dependent on %s, which is to be remediated later", managedPolicy.GetName(), name),
```

controllers/utils/conditions.go
```diff
@@ -73,6 +74,7 @@ var ConditionReasons = struct {
 	PrecacheSpecIncomplete:        "PrecacheSpecIncomplete",
 	PrecacheSpecIsWellFormed:      "PrecacheSpecIsWellFormed",
 	TimedOut:                      "TimedOut",
+	InvalidDependencyOrder:        "InvalidDependencyOrder",
```
jc-rh, line 77: How about `UnresolvableDenpendency`?

PR #578, opened 2023-05-31
OCPBUGS-11769: Identify managed policies with ambiguous names (#556)

irinamihai:
Differentiate between managed policies with the same name in different namespaces:
- determine all the namespaces where we can find managed policies
- if there is at least one managed policy name that can be found in more than one namespace fail the CGU validation and include a condition message that contains the conflicting policy names and namespaces

Cleanup KUTTL child policies (#560)

Description:
- child policies that are not properly cleaned up cause false failure for other KUTTL tests

### DIFF
--- a/controllers/utils/conditions.go
+++ b/controllers/utils/conditions.go
@@ -32,43 +32,45 @@ type ConditionReason string
 
 // ConditionReasons define the different reasons that conditions will be set for
 var ConditionReasons = struct {
-	Completed                  ConditionReason
-	ClusterSelectionCompleted  ConditionReason
-	ValidationCompleted        ConditionReason
-	BackupCompleted            ConditionReason
-	PrecachingCompleted        ConditionReason
-	Failed                     ConditionReason
-	IncompleteBlockingCR       ConditionReason
-	InProgress                 ConditionReason
-	InvalidPlatformImage       ConditionReason
-	MissingBlockingCR          ConditionReason
-	NotAllManagedPoliciesExist ConditionReason
-	NotEnabled                 ConditionReason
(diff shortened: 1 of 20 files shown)


controllers/utils/conditions.go
```diff
@@ -32,43 +32,45 @@ type ConditionReason string
...
+	ValidationCompleted:           "ValidationCompleted",
+	BackupCompleted:               "BackupCompleted",
+	PrecachingCompleted:           "PrecachingCompleted",
+	Failed:                        "Failed",
+	IncompleteBlockingCR:          "IncompleteBlockingCR",
+	InProgress:                    "InProgress",
+	InvalidPlatformImage:          "InvalidPlatformImage",
+	MissingBlockingCR:             "MissingBlockingCR",
+	NotAllManagedPoliciesExist:    "NotAllManagedPoliciesExist",
+	AmbiguousManagedPoliciesNames: "AmbiguousManagedPoliciesNames",
```
jc-rh, line 66: Should it be `AmbiguousManagedPolicyNames`?

irinamihai: The initial commit has it the same as in here. I did a cherry-pick without changing: https://github.com/openshift-kni/cluster-group-upgrades-operator/commit/499ad67a462aade819b70cfbfb98e7b56be02dc3#diff-9dc0c6962ba7ab03a8510d5d877f487287f4517b59d757ddd02544008a6dcb74R46

jc-rh: Right, we'll have to do another PR in main anyways if we decide to update the name.

PR #580, opened 2023-05-31
CNF-8526: Update TALM to handle new PreCachingConfig CR

sakhoury:
This commit contains the following changes:
- changes for TALM to accept and process the new PreCachingConfig CR associated with a CGU for the pre-caching job.
- changes to support parsing the spaceRequired parameter specified in either base 2 or base 10 string format (e.g. "40 GB" or "40 GiB" respectively).
- addition of spaceRequired field to the precache-spec-cm-create template.
- unit tests and kuttl tests to verify functioning of added logic.

Note: deprecation messages are logged wherever overrides are taken from the cluster-group-upgrade-overrides ConfigMap.

/cc @jc-rh @sudomakeinstall2 @vitus133 @irinamihai


controllers/preCachingConfig.go
```diff
@@ -0,0 +1,80 @@
...
+		r.Log.Info("getPreCachingConfigSpec: no preCachingConfig CR specified")
+		return preCachingConfig, nil
+	}
+
+	err := r.Get(ctx, types.NamespacedName{Namespace: preCachingConfigNamespace,
+		Name: preCachingConfigName}, preCachingConfig)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return preCachingConfig, nil
```
jc-rh, line 37: Shouldn't this be treated as an error?

sakhoury: My initial implementation did treat the `IsNotFound` as an error, however, I then decided to follow the same process as that for retrieving the cluster-group-upgrade-overrides ConfigMap, ref: https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/controllers/overrides.go#L36-L40. 

I can change this back to return the error instead. Please confirm.

jc-rh: Config map wasn't a feature published in the API. Its name is fixed and it's expected that it doesn't always exist. It's totally different from the new config CR that is specified by customer in the actual CGU spec.

sakhoury: Very good point! In that case the not found resource should be reported as an error! I will change this.

controllers/preCachingConfig.go
```diff
@@ -0,0 +1,76 @@
...
+	if preCachingConfigName == "" {
+		r.Log.Info("getPreCachingConfigSpec: no preCachingConfig CR specified")
+		return preCachingConfig, nil
+	}
+
+	err := r.Get(ctx, types.NamespacedName{Namespace: preCachingConfigNamespace,
+		Name: preCachingConfigName}, preCachingConfig)
+
+	if err != nil {
+		return preCachingConfig, err
```
jc-rh, line 35: Not sure if the err print will look good in the condition. Might be a good idea to wrap the error with additional text. You need to manually test e2e at least once if you don't want to add new step and it should look like this:

```
  - message: "Precaching spec is incomplete: inconsistent precaching configuration: olm index provided, but no packages"
    reason: PrecacheSpecIncomplete
    status: "False"
    type: PrecacheSpecValid
```

sakhoury: Done! A new TestStep has been added to `tests/kuttl/tests/pre-caching-complete-with-configs` to verify this behaviour.

controllers/precacheFsm.go
```diff
@@ -67,9 +67,17 @@ func (r *ClusterGroupUpgradeReconciler) precachingFsm(ctx context.Context,
...
+		spec, err = r.includePreCachingConfigs(ctx, clusterGroupUpgrade, &spec)
 		if err != nil {
-			return err
+			r.Log.Error(err, "Failed to get PreCachingConfig resource")
+			utils.SetStatusCondition(
+				&clusterGroupUpgrade.Status.Conditions,
+				utils.ConditionTypes.PrecacheSpecValid,
+				utils.ConditionReasons.PrecacheSpecIncomplete,
+				metav1.ConditionFalse,
+				"Precaching spec is incomplete: inconsistent PreCachingConfig object",
```
jc-rh, line 78: This generic error message isn't very helpful.  Why not include the err in the message, similar to the next if block?

vitus133 (commented):
Hey, impressive work!
Added a small comment, otherwise looks good :)

controllers/precache.go
```diff
@@ -225,30 +225,102 @@ func (r *ClusterGroupUpgradeReconciler) deployDependencies(
...
 		}
 	}
-	return image, nil
+	return preCacheImage, nil
+}
+
+// parseSpaceRequired parses the spaceRequired string value (can be a floating-point value) and converts it to
+// an integer string value representing the space required on the managed cluster in Gigabytes.
+// Note that the parsed value is rounded up to the nearest integer via the math.Ceil function.
+func parseSpaceRequired(spaceRequired string) (string, error) {
```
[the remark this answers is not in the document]
sakhoury: Thank you for the suggestion. I have applied the [github.com/docker/go-units](https://github.com/docker/go-units) package for parsing the `spaceRequired` parameter.

sudomakeinstall2:
Great work, Sharat!

PR #586, opened 2023-06-02
OCPBUGS-7389: pre-cache container mounts host's root

irinamihai:
Description:
- mount only the needed host paths for the pre-cache container with the needed privilege access
- change the check_space script to make the check for host instead of the pre-cache container
- update tests

### DIFF
--- a/pre-cache/precache.sh
+++ b/pre-cache/precache.sh
@@ -1,12 +1,22 @@
-#!/bin/bash
+#!/usr/bin/bash
 
 set -e
 
+# Setup the environment for running podman.
+mkdir /host/dev
+mknod -m 0666 /host/dev/null c 1 3
+mkdir /host/dev/shm
+/host/usr/bin/mount -t tmpfs -o rw,nosuid,nodev,noexec,relatime tmpfs /host/dev/shm
+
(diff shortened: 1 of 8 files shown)


vitus133:
Impressive grinding!
/approve

pre-cache/precache.sh
```diff
@@ -1,12 +1,22 @@
-#!/bin/bash
+#!/usr/bin/bash
 
 set -e
 
+# Setup the environment for running podman.
+mkdir /host/dev
+mknod -m 0666 /host/dev/null c 1 3
+mkdir /host/dev/shm
+/host/usr/bin/mount -t tmpfs -o rw,nosuid,nodev,noexec,relatime,size=2G tmpfs /host/dev/shm
```
rrendec, line 9: You should not specify the tmpfs size explicitly. It defaults to half the size of the RAM, and can be viewed using `df`. See https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html for details. On a regular system, `systemd` seems to do the same thing for `/dev/shm` (not specify the size explicitly).

rrendec (approved):
LGTM with the latest changes

PR #589, opened 2023-06-09
CNF-8527: Update pre-cache scripts to pull additional images

sakhoury:
This commit contains the following changes:
- refactor pre-cache/common script with more logging options.
- refactor pre-cache/pull script to pull platform images and additional user-specified images.
- updated the "spaceRequired" field in controllers/templates/precache-templates.go to reflect string output.
- updated the expected spaceRequired fields in the associated kuttl tests as string values.

### DIFF
--- a/controllers/templates/precache-templates.go
+++ b/controllers/templates/precache-templates.go
@@ -60,7 +60,7 @@ spec:
         additionalImages: |{{ range .AdditionalImages }}
           {{ . }} {{ end }}
         platform.image: {{ .PlatformImage }}
-        spaceRequired: {{ .SpaceRequired }}
+        spaceRequired: "{{ .SpaceRequired }}"
       kind: ConfigMap
       metadata:
         name: pre-cache-spec
--- a/pre-cache/pull
+++ b/pre-cache/pull
@@ -3,19 +3,34 @@
 cwd="${cwd:-/tmp/precache}"
 . $cwd/common
 
-mirror_images() {
+wait_image(){
+    local img=$1
+    local pid=$2
 
-    if ! [[ -f $pull_spec_file ]]; then
-        log_debug "no pull spec provided - misconfiguration error"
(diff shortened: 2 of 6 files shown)


pre-cache/pull
```diff
@@ -27,59 +42,88 @@ mirror_images() {
...
-            fi
+            wait_image $pid ${pids[$pid]}
           done
           # Once the batch is processed, reset the new batch size and clear the processes hash for the next one
           max_bg=$max_pull_threads
           pids=()
         fi
     done
+
+    # Make sure that all pull-threads have been processed
```
jc-rh, line 61: Why do we need to redo this?

controllers/templates/precache-templates.go
jc-rh, line 1: Did you mean to remove line 144-145 as well?

pre-cache/pull
sakhoury: This is required especially for cases when the number of images to be pulled are less than the batch size (`max_bg`). Without the added logic, assuming the pulls commence without any failures, the script proceeds normally to check for failed pulls and then terminates if no pulls have failed returning a success. With termination of the script - the Kubernetes job also terminates - resulting in termination of the background pull threads. If the image size is large - the pull thread will not finish in time (i.e. end of script) thus pre-caching will pre-emptively fail and incorrectly report a success. I observed this with manual tests involving 3 additional-images, 1 of which was close to a Gigabyte.
Thus, it is necessary to introduce a wait on background threads before proceeding to retrying failed images.

controllers/templates/precache-templates.go
sakhoury: This is removed in a subsequent PR which involves checking disk space.

pre-cache/pull
sakhoury: Also worthy to note that when the number of image to be pulled < `max_bg`, then the first `wait_image` loop is not executed.

PR #606, opened 2023-07-07
Operator upgrade MCV/MCA cleanup

jc-rh:


controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,74 @@ func DeleteMultiCloudObjects(
 	}
 
 	for _, mcv := range mcvList.Items {
-		multiCloudLog.Info("[DeleteMultiClusterObjects] Delete ManagedClusterView", "name", mcv.Name)
+		multiCloudLog.Info("[DeleteMultiClusterObjects] Delete ManagedClusterView", "ns", mcv.Namespace, "name", mcv.Name)
```
irinamihai, line 321: [DeleteManagedClusterViews] instead of [DeleteMultiClusterObjects]

controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,74 @@ func DeleteMultiCloudObjects(
...
+		client.MatchingLabels(labels),
+	}
+
+	mcaList := &actionv1beta1.ManagedClusterActionList{}
+	if err := c.List(ctx, mcaList, opts...); err != nil {
+		return err
+	}
+
+	for _, mca := range mcaList.Items {
+		multiCloudLog.Info("[DeleteMultiClusterObjects] Delete ManagedClusterAction", "ns", mca.Namespace, "name", mca.Name)
```
irinamihai, line 346: [DeleteManagedClusterActions] instead of [DeleteMultiClusterObjects]

controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,74 @@ func DeleteMultiCloudObjects(
...
 }
 
-// DeleteAllMultiCloudObjects cleans up views associated to all clusters in the list.
-func DeleteAllMultiCloudObjects(
-	ctx context.Context, c client.Client, clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, allClustersForUpgrade []string) error {
-	// Only need to delete ManagedClusterView as ManagedClusterActions get deleted automatically.
-	for _, clusterName := range allClustersForUpgrade {
-		err := DeleteMultiCloudObjects(ctx, c, clusterGroupUpgrade, clusterName)
-		if err != nil {
+// DeleteManagedClusterActions cleans up actions associated to a cluster.
```
irinamihai, line 329: I would expand this and add the explanation that although MCAs get automatically deleted after 60 seconds, they can get stuck if the managed cluster becomes unavailable.

controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,74 @@ func DeleteMultiCloudObjects(
...
+	for _, mca := range mcaList.Items {
+		multiCloudLog.Info("[DeleteMultiClusterObjects] Delete ManagedClusterAction", "ns", mca.Namespace, "name", mca.Name)
+		if err := c.Delete(ctx, &mca); err != nil {
 			return err
 		}
 	}
+	return nil
+}
+
+// FinalMultiCloudObjectCleanup cleans up views associated to all unsuccessful clusters from the current batch if known
```
irinamihai, line 354:
```suggestion
// FinalMultiCloudObjectCleanup cleans up views associated to all unsuccessful clusters from the current batch, if known
```

controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,74 @@ func DeleteMultiCloudObjects(
...
+	// Only need to cleanup for clusters that are not completed
+	succeededClusters := make(map[string]bool)
+	for _, clusterState := range clusterGroupUpgrade.Status.Clusters {
+		if clusterState.State == ClusterRemediationComplete {
+			succeededClusters[clusterState.Name] = true
+		}
+	}
+
+	// No current batch, clean up all just in case
+	if clusterGroupUpgrade.Status.Status.CurrentBatch == 0 {
```
irinamihai, line 366: Hmm - how would this work when the CGU hasn't been enabled?

jc-rh: It would be the same as before where we try cleaning up for every cluster all the time. I suppose we could add more conditional checks to further optimize the CGU delete case.

irinamihai: I was thinking about the case where we're deleting the CGU without /before enabling it. Would that work or would it crash here because clusterGroupUpgrade.Status.Status is not defined?

jc-rh: It wouldn't crash. If plan is populated, we clean up all. If not, nothing to do. Anyway, it's all changed now. Please review the updated logic.

controllers/utils/multicloud_util.go
```diff
@@ -302,25 +318,75 @@ func DeleteMultiCloudObjects(
...
+	for _, mca := range mcaList.Items {
+		multiCloudLog.Info("[DeleteManagedClusterActions] Delete ManagedClusterAction", "ns", mca.Namespace, "name", mca.Name)
+		if err := c.Delete(ctx, &mca); err != nil {
 			return err
 		}
 	}
+	return nil
+}
+
+// FinalMultiCloudObjectCleanup cleans up views associated to all unsuccessful clusters from the current batch, if known.
```
irinamihai, line 356:
```suggestion
// FinalMultiCloudObjectCleanup cleans up views associated to all clusters from the current batch that haven't completed, if known.
```

jc-rh: Updated

PR #613, opened 2023-07-19
Add multiple worker threads

jc-rh:
Needed for large scale ZTP deployment with higher number of policies and hub side templatization. Peak CPU use goes up to .5 core from .1 core on the hub. Also provides better responsiveness in general when multiple CGUs are being created.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -38,6 +40,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -2112,6 +2115,19 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Version: "v1",
 	})
 
+	maxConcurrency, set := os.LookupEnv("TALM_CGU_CTRL_WORKER_COUNT")
+	var maxConcurrentReconciles int
+	var err error
+	if set {
+		maxConcurrentReconciles, err = strconv.Atoi(maxConcurrency)
+		if err != nil {
+			r.Log.Info("Invalid value %s for TALM_CGU_CTRL_WORKER_COUNT, using the default: %i", maxConcurrency, utils.DefaultCGUControllerWorkerCount)
+			maxConcurrentReconciles = utils.DefaultCGUControllerWorkerCount
+		}
+	} else {
+		maxConcurrentReconciles = utils.DefaultCGUControllerWorkerCount
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ranv1alpha1.ClusterGroupUpgrade{}, builder.WithPredicates(predicate.Funcs{
 			UpdateFunc: func(e event.UpdateEvent) bool {
@@ -2138,5 +2154,7 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 			CreateFunc:  func(ce event.CreateEvent) bool { return false },
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
 			DeleteFunc:  func(de event.DeleteEvent) bool { return false },
-		})).Complete(r)
+		})).
+		WithOptions(controller.Options{MaxConcurrentReconciles: maxConcurrentReconciles}).
+		Complete(r)
 }
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -2130,19 +2130,6 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Version: "v1",
 	})
 
-	maxConcurrency, set := os.LookupEnv(utils.CGUControllerWorkerCountEnv)
-	var maxConcurrentReconciles int
-	var err error
-	if set {
-		maxConcurrentReconciles, err = strconv.Atoi(maxConcurrency)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -2112,6 +2115,19 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Version: "v1",
 	})
 
+	maxConcurrency, set := os.LookupEnv("TALM_CGU_CTRL_WORKER_COUNT")
```
sakhoury, line 2118: Suggestion: perhaps we should declare the environment variable name in utils/constants.go and reference the constant here rather than defining the value directly. See comment below.

controllers/utils/constants.go
```diff
@@ -2,6 +2,9 @@ package utils
 
 import "k8s.io/apimachinery/pkg/runtime/schema"
 
+// DefaultCGUControllerWorkerCount default number of worker threads for the CGU controller
+const DefaultCGUControllerWorkerCount = 5
```
sakhoury, line 6: Building on my earlier comment, we could bundle all the CGU Controller constants here:
```suggestion
// CGUController constants
const (
	CGUCtrllWorkerCountEnv          = "TALM_CGU_CTRL_WORKER_COUNT"
	DefaultCGUControllerWorkerCount = 5
)

```

sakhoury: I noticed we still reference `talo` in both backup and pre-caching related templates. Perhaps to remain consistent, we should refer to the environment variable as  `"TALO_CGU_CTRL_WORKER_COUNT"`? Either way, I don't mind - I simply wanted to highlight this here.

sakhoury (commented):
Looks good, added a few minor suggestions!

controllers/utils/constants.go
```diff
@@ -2,6 +2,13 @@ package utils
 
 import "k8s.io/apimachinery/pkg/runtime/schema"
 
+// DefaultCGUControllerWorkerCount default number of worker threads for the CGU controller
+// CGUController constants
+const (
+	CGUCtrllWorkerCountEnv          = "TALM_CGU_CTRL_WORKER_COUNT"
```
sakhoury, line 8: To be consistent, either refer to both CGU controller constants as `...CGUControllerWorker...` or `...CGUCtrlWorker...`

controllers/utils/constants.go
```diff
@@ -2,6 +2,13 @@ package utils
 
 import "k8s.io/apimachinery/pkg/runtime/schema"
 
+// DefaultCGUControllerWorkerCount default number of worker threads for the CGU controller
```
sakhoury, line 5: Move code comment down

controllers/clustergroupupgrade_controller.go
```diff
@@ -2112,6 +2115,19 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Version: "v1",
 	})
 
+	maxConcurrency, set := os.LookupEnv(utils.CGUControllerWorkerCountEnv)
+	var maxConcurrentReconciles int
+	var err error
+	if set {
+		maxConcurrentReconciles, err = strconv.Atoi(maxConcurrency)
+		if err != nil {
```
irinamihai, line 2123: Shouldn't we also check for maxConcurrentReconciles to be > 0?

controllers/clustergroupupgrade_controller.go
```diff
@@ -2112,6 +2115,19 @@ func (r *ClusterGroupUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Version: "v1",
 	})
 
+	maxConcurrency, set := os.LookupEnv(utils.CGUControllerWorkerCountEnv)
```
irinamihai, line 2118:
```suggestion
	maxConcurrencyValue, maxConcurrencyValueIsSet := os.LookupEnv(utils.CGUControllerWorkerCountEnv)
```

controllers/clustergroupupgrade_controller.go
jc-rh: Added

controllers/clustergroupupgrade_controller.go
```diff
@@ -2087,6 +2090,21 @@ func (r *ClusterGroupUpgradeReconciler) handleCguFinalizer(
...
+	if isSet {
+		count, err = strconv.Atoi(maxConcurrency)
+		if err != nil || count < 1 {
+			r.Log.Info("Invalid value '%s' for %s, using the default: %i", maxConcurrency, utils.CGUControllerWorkerCountEnv, utils.DefaultCGUControllerWorkerCount)
+			count = utils.DefaultCGUControllerWorkerCount
+		}
+	} else {
+		count = utils.DefaultCGUControllerWorkerCount
+	}
+	return
```
irinamihai, line 2105:
```suggestion
	return count
```

jc-rh: Is this recommended for named return?

irinamihai: Sorry, my bad, pls disregard.

controllers/clustergroupupgrade_controller.go
```diff
@@ -2087,6 +2090,21 @@ func (r *ClusterGroupUpgradeReconciler) handleCguFinalizer(
 	return utils.DontReconcile, nil
 }
 
+func (r *ClusterGroupUpgradeReconciler) getCGUControllerWorkerCount() (count int) {
+	maxConcurrency, isSet := os.LookupEnv(utils.CGUControllerWorkerCountEnv)
+	var err error
+	if isSet {
+		count, err = strconv.Atoi(maxConcurrency)
+		if err != nil || count < 1 {
```
sakhoury, line 2098: Might be worthwhile to also check the upper limit and impose some restrictions in order to avoid resource hogging...

controllers/clustergroupupgrade_controller.go
```diff
@@ -2087,6 +2090,21 @@ func (r *ClusterGroupUpgradeReconciler) handleCguFinalizer(
 	return utils.DontReconcile, nil
 }
 
+func (r *ClusterGroupUpgradeReconciler) getCGUControllerWorkerCount() (count int) {
```
sakhoury, line 2093: Please add a unit test for this function

controllers/clustergroupupgrade_controller.go
jc-rh: This is mainly for internal use, e.g. experimenting in scale lab.

jc-rh: k8s controller code probably handles 0 and negative values already. We are really stretching here to do all these checks just for logging purposes.

irinamihai: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controller.go#L130
True, true, but I think it's still nice to have this value set to what we need instead of the defaults set my controller-runtime :thinking:

jc-rh: Yes. Again, it comes down to who will be setting this and how often. I should have made it clear that it will be really rare.

PR #628, opened 2023-08-03
OCPBUGS-16742: Skip upgrade graph check for templatized cluster version

jc-rh:


controllers/utils/policy_template.go
```diff
@@ -51,8 +51,16 @@ var (
 	//      $2: resource kind
 	//      $3: resource namespace
 	regexpLookup = regexp.MustCompile(`{{hub.*lookup\s+"(.*?)"\s+"(.*?)"\s+"(.*?)".*hub}}`)
+
+	// This expression matches all types templates
```
irinamihai, line 55:
```suggestion
	// This expression matches all template types
```

controllers/validation.go
```diff
@@ -108,7 +116,30 @@ func (r *ClusterGroupUpgradeReconciler) extractOpenshiftImagePlatformFromPolicie
...
 	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, policies []*unstructured.Unstructured) error {
 
-	_, err := r.extractOpenshiftImagePlatformFromPolicies(policies)
+	versionInfo, err := extractOCPVersionInfoFromPolicies(policies)
+
+	if err == nil {
+		if !versionInfo.clusterVersionCRFound || versionInfo.image != "" {
+			return nil
+		}
+		// check for all the required variables needed to make http call and retrieve image
```
irinamihai, line 125:
```suggestion
		// Check for all the required parameters needed to make the update graph HTTP call and retrieve the image
```

controllers/validation.go
```diff
@@ -108,7 +116,30 @@ func (r *ClusterGroupUpgradeReconciler) extractOpenshiftImagePlatformFromPolicie
...
+		if versionInfo.upstream == "" || versionInfo.channel == "" || versionInfo.version == "" {
```
irinamihai, line 126: Shouldn't we also make sure here that versionInfo.image == ""?
If we reach this line here it doesn't necessarily mean that versionInfo.image is empty, right?

controllers/validation.go
```diff
@@ -108,7 +116,30 @@ func (r *ClusterGroupUpgradeReconciler) extractOpenshiftImagePlatformFromPolicie
 func (r *ClusterGroupUpgradeReconciler) validateOpenshiftUpgradeVersion(
 	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, policies []*unstructured.Unstructured) error {
 
-	_, err := r.extractOpenshiftImagePlatformFromPolicies(policies)
+	versionInfo, err := extractOCPVersionInfoFromPolicies(policies)
+
+	if err == nil {
```
irinamihai, line 121: With the current construct, we're not updating the CGU status with the first error - we might actually be getting a couple of errors, but we're only saving the last one in the status. It might be better to move this `if err == nill` in a function and `return err` for each error case? Just an idea.

controllers/validation.go
jc-rh: Yes, it's checked on line 122

controllers/validation.go
jc-rh: Not sure what you mean. If something happens in the extract function, we skip this and set status at the end.

jc-rh: I see what you mean now. The branches in this if block are actually mutually exclusive. Let me update the last else block. The image check there is actually redundant. The query function already does that.

irinamihai: Ah, nevermind - yes, it looks good.

PR #637, opened 2023-08-14
OCPBUGS-17382: Fix for CGU not being autocreated

fontivan:
- If you were to delete the CGU and then remove ztp-done from a node, then the CGU would previously not be auto-created again
- This fixes that behaviour so that it does not matter which operation you do first, the auto-creation should now run again regardless

### DIFF
--- a/controllers/managedclusterForCGU_controller.go
+++ b/controllers/managedclusterForCGU_controller.go
@@ -289,7 +289,15 @@ func (r *ManagedClusterForCguReconciler) SetupWithManager(mgr ctrl.Manager) erro
 				GenericFunc: func(e event.GenericEvent) bool { return false },
 				CreateFunc:  func(e event.CreateEvent) bool { return true },
 				DeleteFunc:  func(e event.DeleteEvent) bool { return false },
-				UpdateFunc:  func(e event.UpdateEvent) bool { return false },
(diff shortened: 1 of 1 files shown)


fontivan:
Manually retested this morning on my redeployed lab and can confirm I am getting my CGU and Label back due to autocreation regardless of which one I deleted first

controllers/managedclusterForCGU_controller.go
```diff
@@ -289,7 +289,15 @@ func (r *ManagedClusterForCguReconciler) SetupWithManager(mgr ctrl.Manager) erro
 				GenericFunc: func(e event.GenericEvent) bool { return false },
 				CreateFunc:  func(e event.CreateEvent) bool { return true },
 				DeleteFunc:  func(e event.DeleteEvent) bool { return false },
-				UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					// Check if the event was deleting the label "ztp-done"
+					// We want to return true for that event only, and false for everything else
+					const labelName = "ztp-done"
```
jc-rh, line 295: already defined on line 52

jc-rh (approved):

PR #647, opened 2023-08-23
Revert "OCPBUGS-17037: Use a more specific label for TALM pod selectors"

fontivan:
Reverts openshift-kni/cluster-group-upgrades-operator#636


fontivan:
I was able to verify a fix before this was merged therefore I am closing the revert and will open the fix shortly

PR #649, opened 2023-08-23
OCPBUGS-17037: Fix upgrade failure due to immutable label selectors

fontivan:
This fixes upgrade due to immutable selector labels by adding a -v2 suffix to the controller manager deployment.

This is the intended solution for when you need to make changes to an immutable field as per OLM: https://github.com/operator-framework/operator-lifecycle-manager/issues/952#issuecomment-639657949

If we need to make another immutable change in the future then we can just increment v2 to v3 and so on.


fontivan:
Screenshot from 2023-08-23 12-46-36
Screenshot from 2023-08-23 12-47-13
Screenshot from 2023-08-23 12-47-19

jc-rh (approved):

PR #658, opened 2023-08-31
OCPBUGS-7387: Refactor backup-agent job container volume and volumeMounts

sakhoury:
This PR contains changes to the backup-agent job container volumes as volumeMounts. Previously, the container mounted the entire root host partition in R/W mode. The volumes and volumeMounts have been scoped down with read-only (RO) access for most mount points.

In order to support the execution of `ostree admin` commands, a new `InitBackup` function has been added to the recovery/launchBackup.go source file.

The expected kuttl test outputs have been updated accordingly.

/cc @jc-rh @donpenney @irinamihai


recovery/cmd/launchBackup.go
```diff
@@ -126,6 +134,61 @@ func LaunchBackup() error {
...
+// InitBackup Prepare backup pre-requisites
+// returns: 			error
+func InitBackup() error {
+
+	hostDev := filepath.Join(host, "dev")
+	hostDevNull := filepath.Join(hostDev, "null")
+	hostSysroot := filepath.Join(host, "sysroot")
+
+	// Create null device (/dev/null) on /host
+	if err := os.MkdirAll(hostDev, 0o644); err != nil {
```
donpenney, line 146: Why 644 for the directory vs 755?

go.mod
```diff
@@ -73,13 +73,13 @@ require (
 	go.uber.org/atomic v1.9.0 // indirect
 	go.uber.org/multierr v1.7.0 // indirect
 	go.uber.org/zap v1.24.0 // indirect
-	golang.org/x/mod v0.10.0 // indirect
-	golang.org/x/net v0.10.0 // indirect
+	golang.org/x/mod v0.12.0 // indirect
```
donpenney, line 76: Looks like there is a separate PR for this upversion, which I think is a more appropriate place for it:
https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/660

Maybe we can get that PR merged first, then rebase against it? I don't see PRs from the dependbot for these other modules. These upversions seem unrelated to the code changes here, so it would be nice to have these done separately

jc-rh: It's merged now.

recovery/cmd/launchBackup.go
sakhoury: The goal was to further restrict the jailed root privileges. I can change to 755.

donpenney: I was just curious, as it seemed unusual for a directory, and didn't match the host /dev permissions. I have no objections, but a comment here would be good to call out the intentional difference

go.mod
jc-rh: Oh, right, the other PR is for x/sys

sakhoury: Should I create a separate PR for `x/net`?

recovery/cmd/launchBackup.go
sakhoury: Changed to match typical host /dev permissions:
```sh
drwxr-xr-x.   19 root root 3460 Aug 23 02:01 dev
```

go.mod
donpenney: That's just my two cents. I prefer to see upversioning done in a separate PR, primarily to isolate the actual code changes, whether for future cherry-pick or readability, or even for association when the upversion is unrelated to the PR. But it's up to @jc-rh . Everything else looks fine to me.

jc-rh: These are indirect dependencies. Sharat mentioned it was updated automatically due to `unix.Mkdev` he used in this fix. I assume that means it's valid to include the upversion in this PR?

donpenney: I'm guessing dependbot hasn't updated these already because of the configuration, that the upversion hasn't been deemed critical and so isn't done automatically?

jc-rh: Not sure why. Maybe it was blocked by other dependbot PRs? We have a few pending ones that require manual updates in our code.

PR #669, opened 2023-09-15
OCPBUGS-18905: Add retry for fetching GCHighThreshold in precaching check_space

sudomakeinstall2:


pre-cache/check_space
```diff
@@ -11,8 +11,25 @@ validate_disk_space() {
...
+      else
+         ((retries=retries+1))
+         if [[ $retries -eq $max_retries ]]; then
+            log_info "Could not get imageGCHighThresholdPercent value from kubelet. Using default value instead."
+            high=85
+         else
+            sleep 5
+        fi
+      fi
+    done
```
sakhoury, line 31: I think it will be beneficial to define a `retry` function in `pre-cache/common` and reference the function here instead. You can then proceed to validate the `high` value after the `retry` function call. The `retry` function is quite a useful utility function and may easily be used in other cases as well. 
This suggestion follows the same procedure applied in `upgrade-recovery.sh` (https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/recovery/bindata/upgrade-recovery.sh#L64-L93).

jc-rh:
@sakhoury Is the first line of this script a typo or a new way of doing shebang?

sakhoury:
> @sakhoury Is the first line of this script a typo or a new way of doing shebang?

@jc-rh good catch! That is a typo indeed! @sudomakeinstall2 can you please change the first line:
```
#!/bin/bash
```

pre-cache/check_space
```diff
@@ -11,8 +11,25 @@ validate_disk_space() {
     SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
```
donpenney, line 11: As a couple of general notes:
- The first line of this file is odd. Is the ./ a typo or some syntax I'm not familiar with?
- This script and a few others in this dir do not have a .sh extension. This means they're bypassing the bash linters that are run as part of the CI - bashate and shellcheck. This probably needs to be a separate action to fix this and address all the warnings

jc-rh: @sudomakeinstall2 Can you raise a story to capture Don's second point?

sudomakeinstall2:
I have addressed the comments. @donpenney @sakhoury @jc-rh

pre-cache/common
```diff
@@ -56,3 +56,34 @@ unmount_index(){
...
+        if [ $rc -eq 0 ]; then
+            log_debug "Command succeeded:" "${cmd[@]}"
+            break
+        fi
+
+        log_error "Command failed:" "${cmd[@]}"
+    done
+
+    return ${rc}
+}
```
donpenney, line 89: Is this indicating a missing EOL character?

sudomakeinstall2: yes, I will add one

pre-cache/check_space
sudomakeinstall2: https://issues.redhat.com/browse/CNF-9789

jc-rh (approved):

PR #670, opened 2023-09-15
Change TALM backend to use selective policy enforcement

Missxiaoguo:
Ref: https://issues.redhat.com/browse/CNF-6505

/cc @jc-rh @sudomakeinstall2


deploy/acm/crds/policy.open-cluster-management.io_placementbindings_crd.yaml
```diff
@@ -23,6 +27,17 @@ spec:
...
             type: string
+          bindingOverrides:
+            description: BindingOverrides defines the overrides to the Subjects
+            properties:
+              remediationAction:
+                description: This field overrides the policy remediationAction on
+                  target clusters
+                enum:
+                - Enforce
+                - enforce
```
irinamihai, line 38: Is this enum really supposed to be "Enforce, enforce" or should it be "inform, enforce"? :thinking:

Missxiaoguo: Yeah, it is "Enforce, enforce".  For now, only overriding to enforce is supported.

pixelsoccupied:
Angie wrote in the ticket that this PR is now ready. Can we merge this?

Missxiaoguo:
/unhold

Rebased and re-tested

pkg/api/clustergroupupgrades/v1alpha1/types.go
```diff
@@ -250,8 +250,6 @@ type ClusterGroupUpgradeStatus struct {
 	PlacementBindings []string `json:"placementBindings,omitempty"`
 	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Placement Rules"
 	PlacementRules []string `json:"placementRules,omitempty"`
-	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Copied Policies"
```
jc-rh, line 253: We can't delete it unless upversion the API. Just mark it as deprecated.

controllers/actions.go
```diff
@@ -133,27 +133,29 @@ func (r *ClusterGroupUpgradeReconciler) deleteClusterLabels(
...
+		}
 	}
-	clusterGroupUpgrade.Status.PlacementRules = nil
 
-	err = utils.DeletePlacementBindings(ctx, r.Client, clusterGroupUpgrade.Namespace, labels)
-	if err != nil {
-		return fmt.Errorf("failed to delete PlacementBindings for CGU %s: %v", clusterGroupUpgrade.Name, err)
-	}
-	clusterGroupUpgrade.Status.PlacementBindings = nil
+	labels := map[string]string{"openshift-cluster-group-upgrades/clusterGroupUpgrade": clusterGroupUpgrade.Name}
```
jc-rh, line 143: What if there are CGUs with the same name in different namespace?

pkg/api/clustergroupupgrades/v1alpha1/types.go
Missxiaoguo: It's a `status` field

controllers/actions.go
jc-rh: FYI, we already have the namespace label for MCV/MCAs

pkg/api/clustergroupupgrades/v1alpha1/types.go
jc-rh: I don't think status fields follow different rules for versioning, do they? They are still part of the API.

PR #672, opened 2023-09-20
Fix pre-cache scripts not getting checked by shellcheck and bashate

sudomakeinstall2:
Add `.sh` extension to pre-caching scripts.
Fix bashate warnings.
Fix shellcheck warnings.

### DIFF
--- a/pre-cache/olm.sh
+++ b/pre-cache/olm.sh
@@ -18,7 +18,7 @@ render_index(){
     index=$1
     packages=$2
     image_mount=$3
-    
+
     if [[ -h "$image_mount/bin/opm" ]]; then
         opm_bin=$image_mount$(readlink "$image_mount/bin/opm")
     else
@@ -26,23 +26,21 @@ render_index(){
     fi
 
     if [[ -d "$image_mount/configs" ]]; then
-      log_debug "Rendering file based catalog"
-      $opm_bin render "$image_mount/configs" > $rendered_index_path
+        log_debug "Rendering file based catalog"
+        $opm_bin render "$image_mount/configs" > $rendered_index_path
     else
-      log_debug "Rendering SQLite catalog"
-      $opm_bin render "$image_mount/database/index.db" > $rendered_index_path
+        log_debug "Rendering SQLite catalog"
+        $opm_bin render "$image_mount/database/index.db" > $rendered_index_path
     fi
     if [[ $? -ne 0 ]]; then
         return 1
     fi
-
 }
 
 extract_packages(){
     local packages
     tr -d " " < $CONFIG_VOLUME_PATH/operators.packagesAndChannels > /tmp/packagesAndChannels
     while IFS= read -r item; do
-    # for item in $(sort -u '/tmp/packagesAndChannels'); do
         pkg=$(echo $item |cut -d ':' -f 1)
         packages="$packages$pkg,"
     done < <(sort -u /tmp/packagesAndChannels)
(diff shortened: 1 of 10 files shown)


pre-cache/olm.sh
```diff
@@ -0,0 +1,99 @@
...
+        log_debug "Operators index is not specified. Operators won't be pre-cached"
+        return 0
+    fi
+    # There could be several indexes, hence the loop
+    sort -u $CONFIG_VOLUME_PATH/operators.indexes | while IFS= read -r index
+    do
+        image_id=$(pull_index $index $PULL_SECRET_PATH)
+        if [[ $? -ne 0 ]]; then
+            log_debug "pull_index failed for index $index"
+            return 1
```
donpenney, line 63: This code is problematic here. The pipe at line 58 creates a subprocess, so this return is actually just returning from that subprocess and exiting the loop, and the function then returns 0. So as written, these failures are lost/ignored.

The loop would need to be rewritten such that the input is a redirect. Something like:

while IFS= read -r index; do
   do_stuff
done < <(sort -u $CONFIG_VOLUME_PATH/operators.indexes)

As an example, try the following script:

#!/bin/bash

function testloop1 {
    echo blah | while read line; do
        echo "In loop1"
        return 1
    done

    echo "Outside loop1"
    return 0
}

function testloop2 {
    while read line; do
        echo "In loop2"
        return 1
    done < <(echo blah)

    echo "Outside loop2"
    return 0
}

testloop1 ; echo "returned: $?"
testloop2 ; echo "returned: $?"

sudomakeinstall2: nice catch.
I am testing the scripts and for some reason reading the file like that does not work in the pod.

this was the original code which worked as expected
```
    for item in $(sort -u '/tmp/packagesAndChannels'); do
        pkg=$(echo $item |cut -d ':' -f 1)
        packages="$packages$pkg,"
    done
```

but this version does not work in the pod.
```
    while IFS= read -r item; do
        pkg=$(echo $item |cut -d ':' -f 1)
        packages="$packages$pkg,"
    done < <(sort -u /tmp/packagesAndChannels)
```
I see this error `/dev/fd/62: No such file or directory`

donpenney: That's the code from the extract_packages function above. Do you mean that function isn't working as written in this update?

pre-cache/olm.sh
```diff
@@ -0,0 +1,99 @@
...
+    if [[ $? -ne 0 ]]; then
+        return 1
+    fi
+}
+
+extract_packages(){
+    local packages
+    tr -d " " < $CONFIG_VOLUME_PATH/operators.packagesAndChannels > /tmp/packagesAndChannels
+    while IFS= read -r item; do
+    # for item in $(sort -u '/tmp/packagesAndChannels'); do
```
donpenney, line 44: Delete the commented-out line

pre-cache/olm.sh
sudomakeinstall2: yes, it is not working

donpenney: I'm not familiar with the content or format on these files, so it's hard for me to comment on how they're being processed. Some comment blocks in the code would be useful here to show an example snippet of the file content, which would help explain the parsing. One suggestion in this function would be to sort the file content once, reading into an array, and then using that array in the for-loop, rather than sorting it multiple times.

For the extract packages, for example, it could be something like:
```
extract_packages(){
    local packages
    local item
    mapfile -t items < <(sed 's/ //g' "${CONFIG_VOLUME_PATH}/operators.packagesAndChannels" | sort -u)
    for item in "${items[@]}"; do
        packages="${packages}${item#*:},"
    done
    echo "${packages%,}"
}
```

sudomakeinstall2: It is wierd but I tried your snipeet and it also fails with the same error.

the file format is like this:
```
cat /host/tmp/packagesAndChannels 
ptp-operator:4.9
sriov-network-operator:4.9
performance-addon-operator:4.9
```

sudomakeinstall2: I tried your snippet it fails with the same error.
We use `chroot` to run this script. The error is related to that.

donpenney: I guess it doesn't like that form of input redirect for some reason. Maybe populating an array and disable this shellcheck warning:
```
extract_packages(){
    local packages
    local item

    # shellcheck disable=SC2207
    items=( $(sed 's/ //g' "${CONFIG_VOLUME_PATH}/operators.packagesAndChannels" | sort -u) )
    for item in "${items[@]}"; do
        packages="${packages}${item#*:},"
    done
    echo "${packages%,}"
}
```

sudomakeinstall2: ok it was fixed by
```
ln -s /proc/self/fd /dev/fd
```

sudomakeinstall2:
@donpenney I addressed your comments. please review again.

PR #698, opened 2023-10-19
OCPBUGS-21749: Fix failing upgrade-recovery backup job 

sakhoury:
This PR addresses the failing upgrade-recovery backup Kubernetes job by adding the necessary mounts to support running podman in lieu of the changes made to the etcdctl dl (ref: https://github.com/openshift/cluster-etcd-operator/commit/45e54425b4fc945136f37eb36e860b076e0a07dd).

Note that by mounting `/run` directory, the `/run/ostree-booted` flag file causes issues when also mounted with other podman related dependencies and `/sysroot`. Thus, the `/run/ostree-booted` flag file is temporarily renamed to `/run/ostree-booted.tmp` to complete the backup process. This file is renamed to the original value before exiting the `LaunchBackup()` function.

/cc @jc-rh @donpenney

### DIFF
--- a/recovery/cmd/launchBackup.go
+++ b/recovery/cmd/launchBackup.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os/exec"
 	"syscall"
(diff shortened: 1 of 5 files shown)


recovery/cmd/launchBackup.go
```diff
@@ -181,6 +197,22 @@ func InitBackup() error {
 		return err
 	}
 
+	// Delete /run/ostree-booted file if empty
+	ostreeBooted := "/run/ostree-booted"
+	if _, err := os.Stat(ostreeBooted); err == nil {
+		if info, err := os.Stat(ostreeBooted); err == nil {
```
irinamihai, line 203: Hmm - why do we need both these lines?

sakhoury: Missed some cleanup :) This has changed...

PR #709, opened 2023-11-07
Update hack/update-codegen.sh since it was deprecated

sudomakeinstall2:
- restructured the api files so that it conforms with new kube_codegen.sh script


pkg/generated/informers/externalversions/clustergroupupgrades/interface.go
sudomakeinstall2, line 1: @jc-rh

jc-rh (approved):

PR #710, opened 2023-11-07
[WIP] OCPBUGS-22811: Remove subpath volume mount usage from TALM

fontivan:
- Subpaths have a lot of weird behaviour that has been causing problem in TALM, NTO, and beyond

### DIFF
--- a/controllers/templates/backup-templates.go
+++ b/controllers/templates/backup-templates.go
@@ -119,51 +119,41 @@ spec:
                   - mountPath: /host/tmp
                     name: host-tmp
                   - mountPath: /host/usr/bin
-                    name: host-usr
+                    name: host-usr-bin
                     readOnly: true
-                    subPath: bin
                   - mountPath: /host/usr/lib
-                    name: host-usr
+                    name: host-usr-lib
                     readOnly: true
-                    subPath: lib
                   - mountPath: /host/usr/lib64
-                    name: host-usr
+                    name: host-usr-lib64
                     readOnly: true
-                    subPath: lib64
                   - mountPath: /host/usr/libexec
-                    name: host-usr
+                    name: host-usr-libexec
                     readOnly: true
-                    subPath: libexec
-                  - mountPath: /host/usr/share/containers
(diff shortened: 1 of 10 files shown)


fontivan:
Based on manual testing this seems to resolve the backup size issues. With TALM from the operator hub it reports a backup size of 27gb but with this fix it reports only 2gb

fontivan:
I was finally able to confirm that precaching and backup both work as expected and that the backup size is more reasonable
image

controllers/templates/backup-templates.go
```diff
@@ -102,128 +102,158 @@ spec:
...
+                  type: Directory
+                name: host-sysroot
+              - hostPath:
+                  path: /tmp
+                  type: Directory
+                name: host-tmp
               - hostPath:
                   path: /usr
                   type: Directory
                 name: host-usr
```
irinamihai, line 196: Do we still need this?

controllers/templates/backup-templates.go
```diff
@@ -102,128 +102,158 @@ spec:
...
+                  type: Directory
+                name: host-usr-local
+              - hostPath:
+                  path: /usr/share/containers
+                  type: Directory
+                name: host-usr-share-containers
               - hostPath:
                   path: /var/
                   type: Directory
                 name: host-var
```
irinamihai, line 224: Same, is this still needed?

controllers/templates/backup-templates.go
fontivan: It looks like we do, I have tried to remove these top level mounts and everything falls apart very quickly

fontivan: I didn't realize the two pods were using different methods to mount /etc so when I tried to remove the high level mounts from both pods that caused my problems. The backup pod does mount /etc but the precache pod just mounts a few files under /etc instead

fontivan:
@sudomakeinstall2  @jc-rh is one of you able to /approve instead?

jc-rh (approved):

PR #718, opened 2023-11-27
Update codegen so it doesn't alter go.mod

jc-rh:


Makefile
```diff
@@ -274,6 +274,8 @@ bundle-push: ## Push the bundle image.
 
 .PHONY: bundle-check
 bundle-check: bundle
+# Workaround for CI which adds phatom dependencies to go.sum
```
sudomakeinstall2, line 277:
```suggestion
# Workaround for CI which adds phantom dependencies to go.sum
```

sudomakeinstall2 (approved):

PR #721, opened 2023-11-28
OCPBUGS-20204: Invalid ConfigurationPolicy names

irinamihai:
Description:
- In the NewSafeResourceName function, TALM truncates the policy name if needed. It could happen that the policy name is truncated such that the last character is ".". In this situation, the policy name would then be concatenated with "-" and a randomly generated suffix resulting in a policy name that follows this model: "policy_name.-suffix". This type of name is not accepted.
- In NewSafeResourceName, when policy names are truncated, make sure the last character of the obtained base string is either '-' or an alphanumerical character.

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -1153,7 +1153,10 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyForCopiedPolicy
 		// Update the metadata name of the ConfigurationPolicy.
 		plcTmplDef := plcTmpl.(map[string]interface{})["objectDefinition"].(map[string]interface{})
 		metadata := plcTmplDef["metadata"]
-		r.updateConfigurationPolicyName(clusterGroupUpgrade, metadata)
+		err := r.updateConfigurationPolicyName(clusterGroupUpgrade, metadata)
+		if err != nil {
+			return err
+		}
 
 		// Ensure the resources referenced in the hub template policy exist if applicable
 		plcTmplDefSpec := plcTmplDef["spec"].(map[string]interface{})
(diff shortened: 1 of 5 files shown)


controllers/utils/common_test.go
```diff
@@ -10,32 +10,50 @@ func TestNewSafeResourceNames(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, MaxPolicyNameLength-1, len(safeName))
+	assert.Equal(t, name[:MaxPolicyNameLength-7]+"-", safeName[:MaxPolicyNameLength-6])
+
+	name = "cgu-sriov-cloudransno-site9-spree-lb-du-cvslcm-4.14.0-rc.4"
+	safeName, err = NewSafeResourceName(name, "", MaxPolicyNameLength, 8)
+	assert.NoError(t, err)
+	assert.Equal(t, MaxPolicyNameLength-9, len(safeName))
+	assert.Equal(t, name[:MaxPolicyNameLength-15]+"-", safeName[:MaxPolicyNameLength-14])
 }
```
sudomakeinstall2, line 59: Is it possible to have a test case for the error scenario?

controllers/clustergroupupgrade_controller.go
```diff
@@ -1193,19 +1196,28 @@ func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyHubTemplate(
...
 func (r *ClusterGroupUpgradeReconciler) updateConfigurationPolicyName(
-	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, metadata interface{}) {
+	clusterGroupUpgrade *ranv1alpha1.ClusterGroupUpgrade, metadata interface{}) error {
 
 	metadataContent := metadata.(map[string]interface{})
 	name := utils.GetResourceName(clusterGroupUpgrade, metadataContent["name"].(string))
-	safeName := utils.GetSafeResourceName(name, clusterGroupUpgrade, utils.MaxPolicyNameLength, 0)
+	safeName, err := utils.GetSafeResourceName(name, clusterGroupUpgrade, utils.MaxPolicyNameLength, 0)
+	if err != nil {
+		return err
```
sudomakeinstall2, line 1205: do you think wrapping these errors is useful?

controllers/utils/common.go
```diff
@@ -56,5 +65,23 @@ func NewSafeResourceName(name, suffix string, maxLength, spareLength int) string
...
+
+	// Make sure base ends in '-' or an alphanumerical character.
+	for !regexp.MustCompile(`^[a-zA-Z0-9-]*$`).MatchString(base[len(base)-1:]) {
+		base = base[:len(base)-1]
+	}
+
+	newSafeResourceName := fmt.Sprintf("%s-%s", base, suffix)
+	// Validate the newSafeResourceName matches the required regex as per
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+	match, err := regexp.MatchString(RFC1123MetadataRegex, newSafeResourceName)
```
jc-rh, line 77: Is this really necessary? We pretty much know base is going to be a well formed name with this change. How can adding the suffix make it not match?

controllers/clustergroupupgrade_controller.go
irinamihai: No need since I removed the error in NewSafeResourceName.

controllers/utils/common_test.go
irinamihai: No need with the latest pushed patch.

controllers/utils/common.go
jc-rh: Sure, please do that if you don't mind. In general, it's kind of bad to include code that can't be tested. The test coverage number will suffer a little :)

jc-rh (approved):

PR #727, opened 2023-12-01
OCPBUGS-22973: Support object-templates-raw for config policies

fontivan:
ACM has support for object-templates-raw that was backported to 2.7, so we should support it in TALM as well

### DIFF
--- a/controllers/utils/policy_template.go
+++ b/controllers/utils/policy_template.go
@@ -61,18 +61,20 @@ func ContainsTemplates(s string) bool {
 	return regexpAllTemplates.MatchString(s)
 }
 
-func stringToYaml(s string) (interface{}, error) {
+// StringToYaml takes a string and attempts to unmarshal it into a YAML
+func StringToYaml(s string) (interface{}, error) {
 	var yamlObj interface{}
 	if err := yaml.Unmarshal([]byte(s), &yamlObj); err != nil {
-		return yamlObj, fmt.Errorf("Could not unmarshal data: %s", err)
+		return yamlObj, fmt.Errorf("could not unmarshal data: %s", err)
 	}
 	return yamlObj, nil
 }
 
-func yamlToString(y interface{}) (string, error) {
+// YamlToString takes an interface and attempts to marshal it into a string
+func YamlToString(y interface{}) (string, error) {
 	b, err := yaml.Marshal(y)
 	if err != nil {
-		return "", fmt.Errorf("Could not marshal data: %s", err)
+		return "", fmt.Errorf("could not marshal data: %s", err)
 	}
 	return string(b), nil
 }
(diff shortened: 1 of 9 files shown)


controllers/utils/policy_template.go
```diff
@@ -61,28 +61,39 @@ func ContainsTemplates(s string) bool {
...
+	var err error
+
+	switch tmpl := tmpl.(type) {
+	case string:
+		tmplStr = tmpl
+	default:
+		tmplStr, err = YamlToString(tmpl)
+		if err != nil {
+			return err
+		}
```
sudomakeinstall2, line 96: can be a function and reused below?

controllers/utils/policy_util.go
```diff
@@ -165,14 +166,34 @@ func InspectPolicyObjects(policy *unstructured.Unstructured) (bool, error) {
...
+			stringTemplate := StripObjectTemplatesRaw(plcTmplDefSpec[ObjectTemplatesRaw].(string))
+
+			var err error
+			configPlcTmpls, err = StringToYaml(stringTemplate)
+			if err != nil {
+				return containsStatus, &PolicyErr{policyName, ConfigPlcFailRawMarshal}
+			}
+
+		default:
+			return containsStatus, &PolicyErr{policyName, ConfigPlcMissAnyObjTmpl}
```
sudomakeinstall2, line 192: can be a function and reused in other files?

sudomakeinstall2:
Changing templates to placeholder can cause errors down the line in the process (precaching version for example), especially since we don't do any validation on these values (precaching might try to pull placeholder:placeholder image). Also the information that TALM wants might be in a template block, and we are removing that template block. So depending on where the templates are some features (I thinkg precahing and approving install plans) might work or might not work. We need to document this somehow and let users know.

controllers/utils/policy_template.go
fontivan: Yep, done in latest commit

fontivan:
> Changing templates to placeholder can cause errors down the line in the process (precaching image and version for example), especially since we don't do any validation on these values (precaching might try to pull placeholder:placeholder image). Also the information that TALM wants might be in a template block, and we are removing that template block. So depending on where the templates are some features (I think precaching and approving install plans) might work or might not work. We can try to see if the information we need from objects are in templates and if so reject the policy, or ignore them and let user know that some feature won't work with templates.

It is possible there could be some problems that come from this as there are effectively infinite use cases for these templates. Though I don't think this will be a problem currently because we are only checking the stripped templates for compliance type and status fields. The actual policies being created in ACM still have all the templating

fontivan:
> I might be wrong but doesn't this function get called on the stripped templates policies?
> 
> https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/e50b230571cc99217ba79c081df0399096ad6087/controllers/precache.go#L124

I can confirm after testing that this is a problem
image

I am not sure we can support precaching with raw templates then since I don't think we have any way to resolve the templating?

sudomakeinstall2:
Yes, I think same goes for subscriptions. 
User can override precahing values with PrecachingCR.
I think we can document that if the user wants to use these feature they should not use templates.

fontivan:
> Yes, I think same goes for subscriptions. User can override precahing values with PrecachingCR though. I think we can document that if the user wants to use these feature they should not use templates.

I had to account for template usage in both precaching and in upgrade validaiton but I think the current solution (most recent commit) will cover all cases

controllers/utils/policy_util.go
```diff
@@ -226,7 +248,62 @@ func UpdateManagedPolicyNamespaceList(policyNs map[string][]string, policyNameAr
...
+	variableUsageRegex := regexp.MustCompile(`{{\s*\$[\w\.]*\s*}}`)
+	variableUsages := variableUsageRegex.FindAllStringSubmatch(result, -1)
+
+	// Get the usage of all of those variables
+	for _, item := range variableUsages {
+		// item[0] is the full usage of a variable
+		// e.g. "{{ $my.example.var }}"
+
+		// Replace all plain usages of variable with placeholder
+		result = strings.ReplaceAll(result, item[0], "placeholder")
```
irinamihai, line 273: Maybe make `placeholder` a constant in utils.constants since it's used quite a lot.

PR #748, opened 2024-02-06
CNF-11189: Add AZTP policy extractor

vitus133:
Monitor ManagedCluster resource for AZTP label
If labelled, extract child policies for the cluster and wrap
the resources in the AZTP configmap.
If labelled for full acceleration, add manifests
for AZTP post-reboot service.
This commit also includes AZTP post reboot service source code
under pkg/aztp. It is built and copied into the TALM image
for post-reboot execution on spokes.
/cc @jc-rh @Missxiaoguo


jc-rh:
Replaced by #767

vitus133:
/reopen 
In case we decide to revert #767

bundle/manifests/cluster-group-upgrades-operator.clusterserviceversion.yaml
```diff
@@ -416,6 +416,8 @@ spec:
                 command:
                 - /manager
                 env:
+                - name: CONTROLLER_IMG
```
rauhersu, line 419: Kindly move the CONTROLLER_IMG right after INSECURE_CALL_GRAPH (as it was done on in.yaml)

PR #751, opened 2024-02-07
OCPBUGS-29208: Fix for ACM Policy Generator Plugin

edcdavid:
For ConfigurationPolicy kind, the name of the object and the inner template are the same when generated with ACM Gen.
In comparison the PGT plugin, adds "-config" to the name of the inner template (see https://github.com/openshift-kni/cnf-features-deploy/blob/dc64c970fa59dacf387f1e432d6a46ea551dbb1c/ztp/policygenerator/policyGen/policyBuilder.go#L422). this leads to overloaded keys in the
clusterGroupUpgrade.Status.SafeResourceNames map.

Changes:
 - namespace+"/"+name key used in the clusterGroupUpgrade.Status.SafeResourceNames map
 - use of namespace instead of namespace length in GetSafeResourceName
 - use of utf8.RuneCountInString() instead of len to support non english characters (see policy web hook at https://github.com/stolostron/governance-policy-propagator/blob/feb9abb241954d7ce64054d7c621b423deece8cf/api/v1/policy_webhook.go#L86-L87 ).
 - adding logs to display the safenames and lenghts

### DIFF
--- a/controllers/utils/common.go
+++ b/controllers/utils/common.go
@@ -3,7 +3,9 @@ package utils
 import (
 	"fmt"
 	"regexp"
+	"unicode/utf8"
 
+	"github.com/go-logr/logr"
 	ranv1alpha1 "github.com/openshift-kni/cluster-group-upgrades-operator/pkg/api/clustergroupupgrades/v1alpha1"
 	"k8s.io/apimachinery/pkg/util/rand"
 )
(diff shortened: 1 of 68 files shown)


controllers/utils/common.go
```diff
@@ -32,25 +34,37 @@ func GetMinOf3(number1, number2, number3 int) int {
...
-		safeName = NewSafeResourceName(name, clusterGroupUpgrade.GetAnnotations()[NameSuffixAnnotation], maxLength, spareLength)
-		clusterGroupUpgrade.Status.SafeResourceNames[name] = safeName
+		safeName = NewSafeResourceName(name, namespace, clusterGroupUpgrade.GetAnnotations()[NameSuffixAnnotation], maxLength, log)
+		clusterGroupUpgrade.Status.SafeResourceNames[namespace+"/"+name] = safeName
+	} else {
+		if log != nil {
+			log.Info("safename already in  clusterGroupUpgrade.Status.SafeResourceNames",
+				"safename", safeName,
+				"namespace", namespace,
+				"safeName+namespace length <= 62", utf8.RuneCountInString(safeName+namespace))
```
jc-rh, line 51: Nit, if we have to keep this log, drop the "<= 62" part.

controllers/utils/common.go
```diff
@@ -59,12 +73,22 @@ func NewSafeResourceName(name, suffix string, maxLength, spareLength int) string
...
 	}
 
 	// The newSafeResourceName should match regex
 	// `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*` as per
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
-	return fmt.Sprintf("%s-%s", base, suffix)
+
+	safename = fmt.Sprintf("%s-%s", base, suffix)
+
+	if log != nil {
```
jc-rh, line 86: I would drop this log for sure.

controllers/utils/constants.go
```diff
@@ -41,7 +41,9 @@ const (
 
 // CR name length limits and suffix annotation
 const (
-	MaxPolicyNameLength    = 63
+	// Maximum length of policy + namespace (not including extra separator dot so 63 -1 = 62)
+	// this is calculated with utf8.RuneCountInString(policy.Name)+utf8.RuneCountInString+utf8.RuneCountInString(policy.Namespace)
+	MaxPolicyNameLength    = 62
```
jc-rh, line 46: I prefer keep the same value here and count the dot in the length calculation. Or you want to rename it to maxPolicyNameLengthExcludingTheDot?

controllers/utils/common.go
edcdavid: fixed

controllers/utils/common.go
edcdavid: fixed

controllers/utils/constants.go
edcdavid: fixed

PR #767, opened 2024-03-20
CNF-11189: Add AZTP binary in a separate image

vitus133:
Monitor ManagedCluster resource for AZTP label
If labelled, extract child policies for the cluster and wrap the resources in the AZTP configmap.
If labelled for full acceleration, add manifests
for AZTP post-reboot service.
This commit also includes AZTP post reboot service source code under pkg/aztp. It is built and copied into a separate image for post-reboot execution on spokes
This PR is an alternative to #748 - either of them can be considered for merging
/cc @jc-rh


controllers/templates/aztp-templlates.go
```diff
@@ -0,0 +1,152 @@
...
+`
+const clusterRole string = `
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ztp-profile-accelerator-clusterrole
+rules:
+# Read cluster version, clusteroperators, configmaps
+# Approve installplans
+# Do anything with OLM resources
```
jc-rh, line 93: These lines are stale?

jc-rh: Maybe go code comment instead and explain that we need everything

controllers/managedclusterForCGU_controller.go
```diff
@@ -86,12 +95,16 @@ func (r *ManagedClusterForCguReconciler) Reconcile(ctx context.Context, req ctrl
...
 	}
 
-	// Stop creating UOCR if ztp of this cluster is done already
+	// Stop creating UOCR or AZTP if ztp of this cluster is done already
 	if _, found := managedCluster.Labels[ztpDoneLabel]; found {
 		r.Log.Info("ZTP for the cluster has completed. "+ztpDoneLabel+" label found.", "Name", managedCluster.Name)
 		return doNotRequeue(), nil
 	}
 
+	if err := r.extractAztpPolicies(ctx, managedCluster); err != nil {
```
jc-rh, line 104: Does it make more sense to do this inside `handleNotReadyCluster`?

jc-rh: Also, not sure why the available condition check was written that way. It could have been 
```
if availableCondition == nil || availableCondition.Status == metav1.ConditionFalse {
    return r.handleNotReadyCluster(managedCluster), nil
} else {
...
}
```

Missxiaoguo: Initially, it had different logs for condition=nil and status=false cases. As they were removed,  the checks can be simply merged.

Missxiaoguo: If it goes into handleNotReadyCluster, it could be either creation case or cluster becomes NotReady again right after it gets reqeued(ubnormal case). 
I am not familiar with the background of this story. Not sure the reason of redo policy extraction? Is it because of the conern where the child policies have not present yet when it initially gets requeued?

vitus133: Yep, I think it makes sense to move it to handleNotReadyCluster. 
For the abnormal case, there is reason, but IMHO also no harm in redoing the policy extraction

vitus133: fixed

jc-rh (approved):

PR #768, opened 2024-03-25
Manifestwork rollout support

jc-rh:
Change summary

- Commit 1: refactoring and API update. No functionality change
- Commit 2: actual implementation for manifestwork rollout
  - Mostly straightforward. Main difference comparing to policy based rollout is that we need to support different versions of the same CR in the list of manifest templates whereas conflicting policies don't make sense.
  - To support this, the progress update in CGU and the resource cleanup in spoke namespace become a bit tricky. Since the completed manifest work has to be deleted right away instead of the end the whole rollout, we make sure the cluster progress in CGU is updated before deleting the completed manifest work and creating the next one in a subsequent reconcile. Otherwise, the progress gets lost if the reconcile with progress update failed for some reason and the completed manifest work was deleted in the same reconcile. With policy based rollout, the controller can rebuild the progress by inspecting the compliant status of each policy.
  - The validate function will be implemented later.
 - Commit 3: kuttl test
   - Success case only. Will add more in subsequent PRs


pixelsoccupied:
/lgtm

good stuff!

PR #780, opened 2024-05-02
IBGU API definition and initial controller

sudomakeinstall2:
ImageBasedGroupUpgrade is used to upgrade a group of clusters using TALM
and IBU.

* Add IBGU API definition
* Add utility functions to create underlying ManifestWorkReplicaSet and
  CGUs
* Add controller to reconcile IBGU and create CGUs
* Gather CGUs statuses and update them in IBGU

/cc @Missxiaoguo @pixelsoccupied @jc-rh @sakhoury 


controllers/imagebasedgroupupgrade_test.go
```diff
@@ -0,0 +1,336 @@
...
+						Namespace: "namespace",
+					},
+					Spec: v1alpha1.ClusterGroupUpgradeSpec{
+						ManifestWorkTemplates: []string{
+							"name-prep",
+						},
+					},
+				},
+				{
+					ObjectMeta: v1.ObjectMeta{
```
jc-rh, line 254: What happens if the append is done before the first CGU completes?

sudomakeinstall2: The next CGU is created but it has the previous CGU as the blocking CGU.

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,97 @@
...
+// RolloutStrategy defines how to rollout ibu
+type RolloutStrategy struct {
+	//kubebuilder:validation:Minimum=1
+	MaxConcurrency int `json:"maxConcurrency"`
+	//+kubebuilder:default=240
+	Timeout int `json:"timeout,omitempty"`
+}
+
+// ImageBasedGroupUpgradeSpec defines the desired state of ImageBasedGroupUpgrade
+// +kubebuilder:validation:XValidation:message="Invalid list of actions", rule="self.actions==['Prep'] || self.actions==['Prep','Upgrade'] || self.actions==['Prep','Upgrade','Finalize'] || self.actions==['Rollback'] || self.actions==['Rollback', 'Finalize'] || self.actions==['Upgrade'] || self.actions==['Finalize']"
```
Missxiaoguo, line 39: Looks like missing self.actions==['Upgrade', 'Finalize']?

It seems like we are not blocking invalid transitions. For example, what happens if the old action is Prep and the new action is Rollback? Or is this handled by the controller?

controllers/imagebasedgroupupgrade_test.go
jc-rh: Great idea. Unfortunately, the blocking feature was never practically usable at scale. The previous CGU has to be 100% successful for the next one to start.

controllers/utils/manifestworkreplicaset.go
```diff
@@ -0,0 +1,227 @@
...
+	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	serializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	mwv1 "open-cluster-management.io/api/work/v1"
+	mwv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
+)
+
+// GenerateAbortManifestWorkReplicaset returns a populated ManifestWorkReplicaSet for prep stage of an ImageBasedUpgrade
```
Missxiaoguo, line 19: nit: abort stage

bundle/manifests/lcm.openshift.io_imagebasedgroupupgrades.yaml
```diff
@@ -0,0 +1,314 @@
...
+              actions:
+                items:
+                  description: ImageBasedUpgradeAction defines the type for the actions
+                    field
+                  enum:
+                  - Abort
+                  - Prep
+                  - Upgrade
+                  - Rollback
+                  - Finalize
```
pixelsoccupied, line 53: Can we have more verbose description here? Basically would be amazing to know how these map to IBU stage?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:generate=true
+
+// ImageBasedUpgradeAction defines the type for the actions field
+// +kubebuilder:validation:Enum=Abort;Prep;Upgrade;Rollback;Finalize
+type ImageBasedUpgradeAction string
```
jc-rh, line 12: Safer to make this a struc for future extension

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,97 @@
...
+// IBUActions defines the string values for valid actions
+var IBUActions = struct {
+	ImageBasedUpgradeAction
+	Prep     ImageBasedUpgradeAction
+	Upgrade  ImageBasedUpgradeAction
+	Rollback ImageBasedUpgradeAction
+	Abort    ImageBasedUpgradeAction
+	Finalize ImageBasedUpgradeAction
+}{
+	Prep:     "Prep",
```
Missxiaoguo, line 23: I remember we talked about making it to a struct for future extension. Is it still true?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,97 @@
...
+	//+kubebuilder:default=240
+	Timeout int `json:"timeout,omitempty"`
+}
+
+// ImageBasedGroupUpgradeSpec defines the desired state of ImageBasedGroupUpgrade
+// +kubebuilder:validation:XValidation:message="Invalid list of actions", rule="self.actions==['Prep'] || self.actions==['Prep','Upgrade'] || self.actions==['Prep','Upgrade','Finalize'] || self.actions==['Rollback'] || self.actions==['Rollback', 'Finalize'] || self.actions==['Upgrade'] || self.actions==['Finalize']"
+type ImageBasedGroupUpgradeSpec struct {
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="IBU Spec"
+	//+kubebuilder:validation:Required
+	IBUSpec lcav1alpha1.ImageBasedUpgradeSpec `json:"ibuSpec"`
```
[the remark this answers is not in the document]
sudomakeinstall2: I specified it as required. Do we still need the pointer?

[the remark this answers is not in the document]
pixelsoccupied: True might be redundant in that case

controllers/utils/manifestworkreplicaset.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	var dst bytes.Buffer
+	s.Encode(ibu, &dst)
+	return dst.Bytes(), nil
+}
+
+func getCGUNameForIBGU(ibgu *ibguv1alpha1.ImageBasedGroupUpgrade, templates []string) string {
+	s := []string{}
+	for _, template := range templates {
+		s = append(s, strings.ReplaceAll(template, ibgu.GetName()+"-", ""))
```
jc-rh, line 192: template name already contains IBGU name, right? it will easily go beyond the limit if ibgu name is repeated many times. Maybe trim it before appending.

controllers/utils/manifestworkreplicaset.go
```diff
@@ -0,0 +1,227 @@
...
+func GenerateClusterGroupUpgradeForIBGU(ibgu *ibguv1alpha1.ImageBasedGroupUpgrade, templateNames, blockingCGUs []string) *ranv1alpha1.ClusterGroupUpgrade {
+	blockingCRs := []ranv1alpha1.BlockingCR{}
+	for _, cguName := range blockingCGUs {
+		blockingCRs = append(blockingCRs, ranv1alpha1.BlockingCR{
+			Name:      cguName,
+			Namespace: ibgu.GetNamespace(),
+		})
+	}
+	enable := true
+	return &ranv1alpha1.ClusterGroupUpgrade{
```
Missxiaoguo, line 207: We need to annotate the cluster with "import.open-cluster-management.io/disable-auto-import" annotation before starting upgrade and remove it after upgrade is done.
```
apiVersion: ran.openshift.io/v1alpha1
kind: ClusterGroupUpgrade
metadata:
  name: cgu-ibu-upgrade
  namespace: default
spec:
  actions:
    beforeEnable:
      addClusterAnnotations:
        import.open-cluster-management.io/disable-auto-import: "true" 
    afterCompletion:
      removeClusterAnnotations:
      - import.open-cluster-management.io/disable-auto-import 
```
If it's abort or rollback, we need to remove the annotation as well.

controllers/imagebasedgroupupgrade.go
```diff
@@ -0,0 +1,287 @@
...
+				oldGeneration := e.ObjectOld.GetGeneration()
+				newGeneration := e.ObjectNew.GetGeneration()
+				// spec update only for SeedGen
+				return oldGeneration != newGeneration
+			},
+			CreateFunc:  func(ce event.CreateEvent) bool { return true },
+			GenericFunc: func(ge event.GenericEvent) bool { return false },
+			DeleteFunc:  func(de event.DeleteEvent) bool { return false },
+		})).
+		Watches(
```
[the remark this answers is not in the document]
sudomakeinstall2: updated

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
sudomakeinstall2: updated.

added validation so that you can only append to the list.

bundle/manifests/lcm.openshift.io_imagebasedgroupupgrades.yaml
sudomakeinstall2: Sure, I am not sure how to add description here?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
sudomakeinstall2: updated

controllers/imagebasedgroupupgrade.go
```diff
@@ -0,0 +1,287 @@
...
+		err := r.Get(ctx, types.NamespacedName{Name: mwrs.Name, Namespace: mwrs.Namespace}, foundMWRS)
+		if err != nil && errors.IsNotFound(err) {
+			r.Log.Info("Creating ManifestWorkReplicaSet", "ManifestWorkReplicaSet", mwrs.Name)
+			err = r.Create(ctx, mwrs)
+			if err != nil {
+				return fmt.Errorf("error creating ManifestWorkReplicaSet: %w", err)
+			}
+		} else if err != nil {
+			return fmt.Errorf("error getting ManifestWorkReplicaSet: %w", err)
+		} else {
```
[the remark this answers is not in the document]
sudomakeinstall2: I don't think so.
If the object exist there would be no errors

controllers/utils/manifestworkreplicaset.go
sudomakeinstall2: updated

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
sudomakeinstall2: updated

controllers/utils/manifestworkreplicaset.go
sudomakeinstall2: this line is doing that. The cgu name only has the ibgu name once

controllers/utils/manifestworkreplicaset.go
sudomakeinstall2: updated

bundle/manifests/lcm.openshift.io_imagebasedgroupupgrades.yaml
pixelsoccupied: Ah so i think it's updating the comment [here](https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/780/files#diff-6bb92e0b8f8550f1b369fc7f4a7ac5eb018424f856c6abd502ffea7cb550e1b8R10-R14) and run make generate should do the trick?

sudomakeinstall2: Oh, I thought you meant add description for each enum item.
updated.

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,104 @@
...
+	// Finalize defines the finalizing stage for image based upgrade
+	Finalize = ImageBasedUpgradeAction{"Finalize"}
+	// Abort defines the aborting stage for image based upgrade
+	Abort = ImageBasedUpgradeAction{"Abort"}
+	// Unknown defines an unknown ImageBasedUpgrade action
+	Unknown = ImageBasedUpgradeAction{""}
+)
+
+// RolloutStrategy defines how to rollout ibu
+type RolloutStrategy struct {
```
jc-rh, line 36: Should we reuse the one defined in CGU?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,104 @@
...
+	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:generate=true
+
+// ImageBasedUpgradeAction defines the type of the action to perform. Abort will stop the upgrade by setting the IBU stage to Idle. Finalize will remove the previous stateroot after upgrade is done by setting the IBU stage to Idle. Prep will start preparing the upgrade by setting the IBU stage to Prep. Upgrade will start the upgrade process by setting the IBU stage to Upgrade. Rollback will pivot back to previous stateroot by setting the IBU stage to Rollback.
+type ImageBasedUpgradeAction struct {
+	// +kubebuilder:validation:Enum=Abort;Prep;Upgrade;Rollback;Finalize
+	Name string `json:"name"`
```
jc-rh, line 13: name sounds like a freeform string. maybe a different field name that implies enum?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,104 @@
...
+	//+kubebuilder:validation:Required
+	RolloutStrategy RolloutStrategy `json:"rolloutStrategy,omitempty"`
+}
+
+// ClusterState defines the current state of a cluster
+type ClusterState struct {
+	Name          string  `json:"name"`
+	State         string  `json:"state"`
+	CurrentAction *string `json:"CurrentAction,omitempty"`
+	Message       *string `json:"message,omitempty"`
```
jc-rh, line 68: Should we bundle the three fields into a currentAction struct? In case we will add cluster level stuff later?

controllers/utils/conditions.go
```diff
@@ -11,13 +11,17 @@ type ConditionType string
...
 	BackupSuceeded     ConditionType
 	ClustersSelected   ConditionType
 	PrecacheSpecValid  ConditionType
 	PrecachingSuceeded ConditionType
 	Progressing        ConditionType
 	Succeeded          ConditionType
 	Validated          ConditionType
+
+	// IBGU
+	ManifestsCreated ConditionType
```
pixelsoccupied, line 24: Does it make sense to have a totally seperate ConditionTypes struct e.g `ConditionTypesIBGU` for ibgu? That way we wont accidentally pull in "invalid" types and just reads clearer?

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
jc-rh: Maybe a better way is to have a list of action status, matching the spec:
```
- name: cluster1
  - action: prep
    state: succeeded
  - action: upgrade
    state: in progress
- name: cluster2
  - action: prep
    state: failed
  - action: upgrade
    state: skipped or not selected
- name: cluster3
  - action: prep
    state: succeeded
  - action: upgrade
    state: succeeded
  - action: finalize
    state: in progress
```

controllers/utils/manifestworkreplicaset.go
```diff
@@ -0,0 +1,240 @@
...
+			Annotations: map[string]string{
+				manifestWorkExpectedValuesAnnotation: expectedValueAnn,
+			},
+		},
+		Spec: mwv1alpha1.ManifestWorkReplicaSetSpec{
+			PlacementRefs: []mwv1alpha1.LocalPlacementReference{
+				{
+					Name: "dummy",
+				},
+			},
```
pixelsoccupied, line 157: Without knowing anything about the ManifestWorkReplicaSetSpec....curious why we need this one LocalPlacementReference placeholder? 

sidenote: maybe `dummy` --> `placeholder`

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
jc-rh: Probably too verbose though

controllers/utils/manifestworkreplicaset.go
jc-rh: If we were to actually populate this programatically, `placeholder` would make sense. In our case, we don't actually use the ManifestWorkReplicaSet controller to rollout the manifests. TALM does it based on the things defined in CGU. That's why I called it dummy in my test files. What's missing here is another ManifestWorkTemplate CRD. Do you think it's better if we define it in our API group instead of abusing ManifestWorkReplicaSet? We will have to decide before making TALM manifestwork support GA.

controllers/utils/manifestworkreplicaset.go
jc-rh: ah, I missed it :+1:

controllers/imagebasedgroupupgrade.go
```diff
@@ -0,0 +1,264 @@
...
+		LabelSelector: labels.SelectorFromSet(labels.Set{utils.CGUOwnerIBGULabel: ibgu.Name}),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to get CGUs for the IBGU: %w", err)
+	}
+	if len(cguList.Items) == 0 {
+		return nil
+	}
+	m := make(map[string]ibguv1alpha1.ClusterState)
+	for _, cgu := range cguList.Items {
```
jc-rh, line 122: The CGUs should be sorted by the order of the actions. Also, we probably can ignore completed CGUs?

controllers/utils/conditions.go
jc-rh: Right, we certainly don't want to copy functions just for this.

controllers/imagebasedgroupupgrade.go
jc-rh: Hmm, I guess we can't skip completed CGUs if they are chained rather than getting enabled by this controller.

controllers/utils/conditions.go
sudomakeinstall2: I will think about some kind of interface for a future PR

controllers/utils/manifestworkreplicaset.go
pixelsoccupied: Ahh just read up on it a bit. So basically if i understanding this correctly...ManifestWorkReplicaSet basically means placementRefs + manifestWorkTemplate

But since we dont use the placementRefs feature (CGU is handling that instead) that just leaves us with only the manifestWork to modify but we need the PlacementRefs is a required field? 


> What's missing here is another ManifestWorkTemplate CRD. Do you think it's better if we define it in our API group instead of abusing ManifestWorkReplicaSet

I really want to say yes. Would that be too much work? So basically we would owns() ManifestWork instead ManifestWorkReplicaSet?

controllers/utils/conditions.go
pixelsoccupied: next PR sounds good to me no problem

controllers/utils/common.go
```diff
@@ -85,3 +88,60 @@ func ContainsTemplates(s string) bool {
...
+		if _, found := mb[x]; !found {
+			diff = append(diff, x)
+		}
+	}
+	return diff
+}
+
+// SortCGUListByIBUAction orders the CGUs by the last action in the name of cgu
+// with following order prep-upgrade-rollback-finalize-abort
+func SortCGUListByIBUAction(cguList *ranv1alpha1.ClusterGroupUpgradeList) {
```
jc-rh, line 109: This will need to be updated when we support partial abort/rollback in the same CR

controllers/utils/manifestworkreplicaset.go
jc-rh: We'll probably call it ManifestWorkTemplate. It would have to be under our lcm api group. But the definition should be mostly imported from OCM, plus the extension we add for the expected values. Not that it's hard. It's just not ideal to have something really similar but not exactly the same as the OCM CRDs. Ideally, we need the OCM manifest work to support expected values natively (side by side with the feedback rules and jsonpath) and expose the template part as a CRD instead of a subsection of the MWRS.

pkg/api/imagebasedgroupupgrades/v1alpha1/types.go
```diff
@@ -0,0 +1,104 @@
...
+	Clusters []string `json:"clusters,omitempty"`
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Cluster Label Selectors",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	ClusterLabelSelectors []metav1.LabelSelector `json:"clusterLabelSelectors,omitempty"`
+	//kubebuilder:validation:Minimum=1
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Actions"
+	// kubebuilder:validation:XValidation:message="Invalid list of actions", rule="self.map(x,x.action)==['Prep'] || self.map(x,x.action)==['Prep','Upgrade'] || self.map(x,x.action)==['Prep','Upgrade','Finalize'] || self.map(x,x.action)==['Rollback'] || self.map(x,x.action)==['Rollback', 'Finalize'] || self.map(x,x.action)==['Upgrade'] || self.map(x,x.action)==['Upgrade','Finalize'] || self.map(x,x.action)==['Finalize'] || self.map(x,x.action)==['Abort'] || self.map(x,x.action)==['Prep', 'Abort'] || self.map(x,x.action)==['Prep', 'Upgrade', 'Rollback'] || self.map(x,x.action)==['Prep', 'Upgrade', 'Rollback', 'Finalize'] || self.map(x,x.action)==['Upgrade','Rollback'] || self.map(x,x.action)==['Upgrade','Rollback','Finalize']"
```
Missxiaoguo, line 54: I just think of the case where stages are done in multiple IBGUs. For example, IBGU1 has only prep stage. Instead of appending to the existing IBGU1, another IBGU2 is created with upgrade stage.  Do we block/support this case? I suppose ideally we would like a single IBGU to manage the full lifecycle of IBU for the same group of clusters?

PR #787, opened 2024-06-20
CNF-13181: Allow partial completion before moving to next CGU

sudomakeinstall2:
Before this change, when multiple CGUs were chained together, the first CGU had to completely finish before the next CGU was allowed to start. This meant that if one cluster failed, the next CGU could not start for the remaining clusters.

Now, by setting the blockingCRCompletionMode annotation to partial, dependent CGUs start for clusters that are compliant with the previous CGUs.

/cc @jc-rh

### DIFF
--- a/controllers/clustergroupupgrade_controller.go
+++ b/controllers/clustergroupupgrade_controller.go
@@ -395,9 +395,6 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			)
 			nextReconcile = requeueWithMediumInterval()
 		} else {
-			// There are no blocking CRs, continue with the upgrade process.
-			// create backup
-
 			err = r.reconcileBackup(ctx, clusterGroupUpgrade, clusters)
 			if err != nil {
 				r.Log.Error(err, "reconcileBackup error")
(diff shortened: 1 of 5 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -1125,13 +1174,16 @@ func (r *ClusterGroupUpgradeReconciler) blockingCRsNotCompleted(ctx context.Cont
 			continue
 		}
 
-		// If we find a blocking CR that does not contain a true succeeded condition then we add it to the list.
-		if !meta.IsStatusConditionTrue(cgu.Status.Conditions, string(utils.ConditionTypes.Succeeded)) {
+		if partialComplete == true {
+			if meta.IsStatusConditionTrue(cgu.Status.Conditions, string(utils.ConditionTypes.Progressing)) {
```
jc-rh, line 1178: Progressing can be false before the CGU starts. I thought we had to use the existence of the succeeded condition.

sudomakeinstall2: how about progressing=false, succeded=false

sudomakeinstall2: updated

controllers/clustergroupupgrade_controller.go
```diff
@@ -446,6 +443,33 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					r.Log.Error(err, "filterNonCompletedClustersInBlockingCRs")
+					return
+				}
+				if len(clusters) == 0 {
+					utils.SetStatusCondition(
+						&clusterGroupUpgrade.Status.Conditions,
+						utils.ConditionTypes.Progressing,
+						utils.ConditionReasons.Completed,
+						metav1.ConditionFalse,
+						"No clusters available for remediation after filtering non completed clusters",
```
Missxiaoguo, line 458: Maybe more clear with this?
```suggestion
"No clusters available for remediation after filtering out non-completed clusters from blocking CGUs",
```

PR #788, opened 2024-06-20
Add IBU permissions for ManifestWork agent 

sudomakeinstall2:
ManifestWork agent by default doesn't have permission to access IBU
resource. At the start of IBGU reconcile add the required permissions on the
clusters.
/cc @jc-rh @Missxiaoguo @pixelsoccupied

### DIFF
--- a/controllers/imagebasedgroupupgrade.go
+++ b/controllers/imagebasedgroupupgrade.go
@@ -145,6 +145,15 @@ func (r *IBGUReconciler) syncStatusWithCGUs(ctx context.Context, ibgu *ibguv1alp
 func (r *IBGUReconciler) ensureManifests(ctx context.Context, ibgu *ibguv1alpha1.ImageBasedGroupUpgrade) error {
 	manifestWorkReplicaSets := []*mwv1alpha1.ManifestWorkReplicaSet{}
 	manifestWorkReplicaSetsNames := []string{}
+
+	permissionsName := "ibu-permissions"
+	mwrs, err := utils.GeneratePermissionsManifestWorkReplicaset(permissionsName, ibgu.GetNamespace())
+	if err != nil {
+		return fmt.Errorf("Error generating manifestworkreplicaset: %w", err)
+	}
+	manifestWorkReplicaSets = append(manifestWorkReplicaSets, mwrs)
+	manifestWorkReplicaSetsNames = append(manifestWorkReplicaSetsNames, permissionsName)
+
 	ibu := &lcav1.ImageBasedUpgrade{
 		ObjectMeta: v1.ObjectMeta{
 			Name: "upgrade",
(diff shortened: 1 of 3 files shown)


controllers/imagebasedgroupupgrade.go
```diff
@@ -142,6 +150,73 @@ func (r *IBGUReconciler) syncStatusWithCGUs(ctx context.Context, ibgu *ibguv1alp
...
+			Namespace: "",
+			Name:      "ibu-permission-set",
+		},
+		Spec: mwv1.ManifestWorkSpec{
+			Workload: mwv1.ManifestsTemplate{
+				Manifests: []mwv1.Manifest{{RawExtension: runtime.RawExtension{Raw: crBytes}}},
+			},
+		},
+	}
+	for _, cluster := range ibgu.Spec.Clusters {
```
jc-rh, line 203: Is it cleaner to create this as a mwrs and just include it in the CGU?

sudomakeinstall2: updated

jc-rh: `clusters` actually doesn't include anything if the cluster selector is used. Looks like the only way is to create a mwrs and include it in all CGUs.

sudomakeinstall2: I was thinking about re using the `getclustersforupgrade` which uses both clusters and labelselector.
But I think the mwrs is better.

controllers/utils/manifestworkreplicaset.go
```diff
@@ -121,7 +122,71 @@ func GeneratePrepManifestWorkReplicaset(name, namespace string, ibu *lcav1alpha1
...
+	}
+	if !isUnversioned && len(gvks) == 1 {
+		cr.TypeMeta = metav1.TypeMeta{
+			Kind:       gvks[0].Kind,
+			APIVersion: gvks[0].GroupVersion().Identifier(),
+		}
+	}
+	var dst bytes.Buffer
+	s.Encode(cr, &dst)
+	return dst.Bytes(), nil
```
pixelsoccupied, line 186: Ah surprised that upstream doesn't provided a simple byte conversion func. Seems like super useful use case. Nothing in the Unstructured Objects lib either yeah?

pixelsoccupied (commented):
Just a few comments but lgtm otherwsie!

controllers/utils/manifestworkreplicaset.go
pixelsoccupied: How about something like this https://github.com/open-cluster-management-io/api/blob/bd6f2229b53cb61c76ebf528c73bc925d8e0f4c8/utils/work/v1/workapplier/workapplier.go#L166-L181 ?

controllers/utils/manifestworkreplicaset.go
```diff
@@ -59,8 +60,8 @@ func GenerateFinalizeManifestWorkReplicaset(name, namespace string, ibu *lcav1al
 }
 
 // GenerateUpgradeManifestWorkReplicaset returns a populated ManifestWorkReplicaSet for upgrade stage of an ImageBasedUpgrade
-func GenerateUpgradeManifestWorkReplicaset(name, namespace string, ibu *lcav1alpha1.ImageBasedUpgrade) (*mwv1alpha1.ManifestWorkReplicaSet, error) {
-	ibu.Spec.Stage = lcav1alpha1.Stages.Upgrade
```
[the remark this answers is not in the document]
sudomakeinstall2: sure, https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/789

controllers/utils/manifestworkreplicaset.go
sudomakeinstall2: this does not work. It doesn't include apiversion, group and kind.

pixelsoccupied: Good to know! Thanks for checking

jc-rh (approved):

PR #808, opened 2024-07-22
IBGU: remove blocking cgu annotations

sudomakeinstall2:
Since we are using label selectors, we no longer need blocking cgu annotations to filter clusters.

### DIFF
--- a/controllers/imagebasedgroupupgrade.go
+++ b/controllers/imagebasedgroupupgrade.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
-	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


controllers/imagebasedgroupupgrade.go
```diff
@@ -305,13 +299,8 @@ func (r *IBGUReconciler) ensureCGUForPlanItem(
 			r.Log.Info("ManifestWorkReplicaSet already exist", "ManifestWorkReplicaSet", mwrs.Name)
 		}
 	}
-	annotations := make(map[string]string)
```
jc-rh, line 308: Please keep this and the function argument down below. I used it for passing the kuttl suffix annotation to CGU

jc-rh (approved):

PR #821, opened 2024-08-27
Upversion to 4.18

sudomakeinstall2:


jc-rh:
Looks like we forgot to upversion rbac proxy here too. No impact to d/s, but still good to keep it consistent.

jc-rh (approved):

PR #823, opened 2024-09-11
IBGU: Add filtering for abort and rollback

sudomakeinstall2:
- Add filtering for abort, cgu only selects managed clusters without upgrade completed label
- Add filtering for rollback, cgu only selects managed clusters with upgrade completed label
- Add validation for plan so that rollback and abort are only allowed in separate ibgu


controllers/utils/manifestworkreplicaset.go
```diff
@@ -261,6 +261,29 @@ func getLabelSelectorForPlanItem(
 ) []metav1.LabelSelector {
 	result := []metav1.LabelSelector{}
 	switch planItem.Actions[0] {
+	case ibguv1alpha1.Abort:
+		for _, m := range currentSelector {
+			upgradeCompletedNotExist := m.DeepCopy()
+			if upgradeCompletedNotExist.MatchExpressions == nil {
```
jc-rh, line 267: Shall we do this outside now that we add filtering almost all the time?

sudomakeinstall2: won't be simplified that much, since for abort on failure we have two selectors

jc-rh (approved):

PR #826, opened 2024-09-16
[release-4.16] backport IBGU

sudomakeinstall2:


fontivan:
> /cc @fontivan @rauhersu can we use golang 1.22 for talm 4.16?

@sudomakeinstall2 I believe we can but we will have to align this with the midstream build. I will discuss with @rauhersu

fontivan:
> > /cc @fontivan @rauhersu can we use golang 1.22 for talm 4.16?
> 
> @sudomakeinstall2 I believe we can but we will have to align this with the midstream build. I will discuss with @rauhersu

@sudomakeinstall2 Midstream should be ready for this now

PR #846, opened 2024-11-14
Replace docker hub image references with projectquay image references

fontivan:
- Docker hub is rate limited
- The projectquay image tracks the upstream docker hub image with CLAIR dependencies added


fontivan:
Docker rate hub limits are causing the CI failures in #845

PR #932, opened 2025-01-28
Update Dockerfile FROM lines to allow Konflux to build using RH images

fontivan:
- By default, use the quay golang builders
- Optionally allow switching to the RH golang builders
- Add the configuration to the tekton pipelines so Konflux will always use the RH golang builders


rauhersu:
Do we need to override the runtime images with a similar approach / another mechanism will be used later to set the images in Dockerfiles?

fontivan:
> Do we need to override the runtime images with a similar approach / another mechanism will be used later to set the images in Dockerfiles?

That is a good question but I do not currently know the answer

rauhersu (approved):
/lgtm

PR #997, opened 2025-03-14
Tekton: Use a single build pipeline reference for all talm builds

fontivan:
- Reduces the duplication in the .tekton pipelines greatly


yanirq:
nice reduction , will this cause any issues with automatic updates for all the tekton files ?  
Will the automation know to identify the build-pipeline new file and update it if needed ? (assuming there will be updates)

fontivan:
> nice reduction , will this cause any issues with automatic updates for all the tekton files ? Will the automation know to identify the build-pipeline new file and update it if needed ? (assuming there will be updates)

As far as I know it still works based on the conversation with the lvm team, though I suppose there is only one way to find out for sure

PR #1037, opened 2025-04-02
Add missing SHA256 digests for Konflux containers images

fontivan:
- golang-builder:v1.23
- rhel-minimal:9.4
- yq:latest


fontivan:
I have intentionally not updated the digests to the latest here, because I want to see if the Konflux/renovate bot will update them without any additional configuration in renovate.json

PR #1038, opened 2025-04-02
Add ManifestWork as bundle dependency

sudomakeinstall2:


jc-rh:
should we include mwrs too?

PR #1060, opened 2025-04-09
OCPBUGS-54738: Fix panic when cluster is not in current batch

sudomakeinstall2:
after all batches are finished, controller goes
through all previous batches to see
if policies are still compliant; in this case some clusters will not be present in
CurrentBatchRemediationProgress and there is no need to check soaking or modify FirstCompliantAt

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -11,7 +11,7 @@ testDirs:
 - tests/kuttl/cgu/
 - tests/kuttl/ibgu/
 crdDir: ./config/crd/bases/
-timeout: 25
+timeout: 40
 parallel: 1
 namespace: default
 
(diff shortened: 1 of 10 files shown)


kuttl-test.yaml
```diff
@@ -11,7 +11,7 @@ testDirs:
 - tests/kuttl/cgu/
 - tests/kuttl/ibgu/
 crdDir: ./config/crd/bases/
-timeout: 25
+timeout: 40
```
jc-rh, line 14: Can this be set on a per test basis?

sudomakeinstall2: it can be set for assert files, but we need it in the step files.

PR #1142, opened 2025-04-28
Update Konflux References

fontivan:
- Include commit from Konflux bot
- Account for new mandatory parameters
- Update to latest task versions for good measure


fontivan:
Supercedes https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/1134

PR #1206, opened 2025-05-02
Conform to new Konflux policy

fontivan:
- OLM Bundle images must not be multi arch


yanirq:
all konflux tests pass.
/approve

fontivan:
> is there a link where this is stated in the documentation? I wonder if this apply to separate container images

Only applies to OLM bundles so I think standalone containers are not affected

fontivan:
> is there a link where this is stated in the documentation? I wonder if this apply to separate container images

The documentation link is https://conforma.dev/docs/ec-policies/release_policy.html#olm__olm_bundle_multi_arch

PR #1295, opened 2025-05-12
[release-4.18] Backport Konflux changes to 4.18

fontivan:


.konflux/overlay/map_images.in.yaml
```diff
@@ -0,0 +1,10 @@
+# Manager
```
rauhersu, line 1: Need to reference talm images instead.

Caveat: additionally, check the csv per backport as this can be prone to errors: for talm, lca < 4.19 we need to deal with the 'kube-rbac-proxy' image as well and map+pin it.

.konflux/overlay/pin_images.in.yaml
```diff
@@ -0,0 +1,20 @@
+# Manager
```
rauhersu, line 1: Ditto for my comment about 'kube-rbac-proxy'.

It might be advisable to check the csv per backport as I am not sure all talm < 4.19 include the aztp image either.

renovate.json
```diff
@@ -0,0 +1,52 @@
...
+            "matchStrings": [
+                "(?<depName>[\\w\\-\\.\\/]+):?(?<currentValue>[\\w\\-\\.]+)?@(?<currentDigest>sha256:[a-f0-9]+)"
+            ],
+            "versioningTemplate": "docker"
+        }
+    ],
+    "gomod": {
+        "enabled": false
+    },
+    "packageRules": [
```
rauhersu, line 21: Do we need this config snippet ? Wondering if our nudge relationships + cm is all we need to manage yaml files under .konflux

.konflux/overlay/pin_images.in.yaml
fontivan: I applied the hold yesterday when I noticed that kube-rbac-proxy was failing conforma, will discuss with you later

renovate.json
fontivan: I'm not sure, but I like how renovate is currently working so I'm going to leave it alone 😅

.konflux/overlay/pin_images.in.yaml
```diff
@@ -0,0 +1,25 @@
...
+  source: quay.io/openshift-kni/cluster-group-upgrades-operator-precache:4.18.0
+  target: quay.io/redhat-user-workloads/telco-5g-tenant/topology-aware-lifecycle-manager-precache-4-18@sha256:b9742dacb72c5fe3223e39c2a2e1d007f163dcb34ea5278baae9c0934a6f0fc2
+#
+# Recovery
+- key: recovery_img
+  source: quay.io/openshift-kni/cluster-group-upgrades-operator-recovery:4.18.0
+  target: quay.io/redhat-user-workloads/telco-5g-tenant/topology-aware-lifecycle-manager-recovery-4-18@sha256:6f9116954825baf1d8d124981ad7c56cd2193904c4e7b663fe625ee000b693e3
+#
+# Kube RBAC Proxy
+- key: kube_rbac
```
fontivan, line 22: do we need a specific key for kube_rbac? @rauhersu

rauhersu: No, as long as the same key is shared between pin and map yaml files, we are good.

PR #2116, opened 2025-06-25
Add network policies for the operator

karampok:
In that version the network policies resources should pass through the bundle. In order to actually being used OLM operator should be version > 0.32


karampok:
I have verified that there are policies installed during integration step
```
❯ k -n openshift-cluster-group-upgrades get networkpolicies
NAME                                  POD-SELECTOR                                                                                             AGE
cluster-group-upgrades-default-deny   <none>                                                                                                   121m
cluster-group-upgrades-operator       app.kubernetes.io/component=talm,app.kubernetes.io/name=talm-operator,control-plane=controller-manager   121m
```

jc-rh (approved):

PR #2300, opened 2025-07-08
OCPBUGS-58372: Fix TALM performing "afterCompletion" while blocked

sudomakeinstall2:
If cluster is not in the remediationPlan (i.e. already fully compliant with all the policies), TALM will perform afterCompletion actions while it is blocked by another CGU. This commit will fix this by not peforming afterCompltion actions while CGU is blocked.

Performing afterCompletion actions is removed from buildRemediationPlan and it is performed by separate function. So we will only perform afterCompletion actions for compliant clusters after we process blocked CGUs.

### DIFF
--- a/.konflux/overlay/pin_images.in.yaml
+++ b/.konflux/overlay/pin_images.in.yaml
@@ -1,17 +1,17 @@
 # Manager
 - key: manager
   source: quay.io/openshift-kni/cluster-group-upgrades-operator:4.20.0
-  target: quay.io/redhat-user-workloads/telco-5g-tenant/topology-aware-lifecycle-manager-4-20@sha256:2cd8cd83a62ed2c221f055b832e898aed4e8044831bc8b537edfbc2082fa324e
+  target: quay.io/redhat-user-workloads/telco-5g-tenant/topology-aware-lifecycle-manager-4-20@sha256:bfcff2bacd2dec985215ec6ef1a3a59cb77dfc5f4093eeffb38030398cc4b32a
 #
 # Aztp
 - key: aztp_img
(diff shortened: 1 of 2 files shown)


controllers/clustergroupupgrade_controller.go
```diff
@@ -505,7 +506,9 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 			}
 
 			// Rebuild remediation plan since we are about to start the upgrade and want to make sure the non-successful clusters were filtered out
-			r.buildRemediationPlan(ctx, clusterGroupUpgrade, clusters, managedPoliciesInfo.presentPolicies)
+			r.buildRemediationPlan(clusterGroupUpgrade, clusters, managedPoliciesInfo.presentPolicies)
+			r.performAfterCompletionActionsForCompliantClusters(
```
jc-rh, line 510: how about making the function above return the list of compliant clusters?

jc-rh: I meant the fully compliant cluster, i.e. the ones not in the plan

deploy/upgrades/blocking-mechanisms/cgu-a.yaml
```diff
@@ -6,14 +6,20 @@ metadata:
   annotations:
     cluster-group-upgrades-operator/name-suffix: kuttl  
 spec:
+  actions:
+    afterCompletion:
+      deleteClusterLabels:
+        upgrade: ""
```
jc-rh, line 12: Not sure how to verify if the finish label is only added in the last step. For the deletion though, we could verify that  it exists on the clusters until the end?

jc-rh (approved):

PR #2880, opened 2025-07-31
Update project to use new shared script library

fontivan:
- Add telco5g-konflux as a submodule
- Update tekton pipelines to rebuild on library change
- Update makefile to call the library helpers for various Konflux maintenance tasks
- Add yamllint to project and fix yamllint warnings
- Add yq-sort-and-format to makefile
- Delete now redundent hack scripts


rauhersu:
/lgtm
/approve

We might start a thread per operator in the corresponding Slack channel, for folks to be aware of the new submodule way of working. Cc: @jc-rh

PR #3315, opened 2025-08-19
Fix the apiVersion for IDMS files

fontivan:
- According to docs it should be `config.openshift.io/v1` (https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/config_apis/imagedigestmirrorset-config-openshift-io-v1)


.konflux/catalog/fbc-images-resolvable-integration-test-idms.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 # We use this IDMS to check if the images in the FBC are resolvable using the staging or production catalog.
 # https://github.com/konflux-ci/community-catalog/tree/development/pipelines/validate-fbc-images-resolvable
-apiVersion: operator.openshift.io/v1
```
rauhersu, line 4: Thanks for including these changes in o-cloud too

PR #10086, opened 2026-06-29
Emit CGU events.k8s.io/v1 manually using the reconciler's client

greyerof:
The existing EventRecorder (core/v1) is deprecated. Its events.k8s.io/v1 replacement does not support setting annotations, which are essential for CGU event metadata. Since the EventSink interface offers no advantage over using the client directly, the EventRecorder is replaced with a custom Emitter that creates events.k8s.io/v1 Event resources through the reconciler's client, giving full control over annotations, action, and related-object fields.

The Emitter produces one unique Event per lifecycle transition (no dedup/rate-limiting), which matches CGU's low-volume, distinct event model.

Events created through events.k8s.io/v1 remain backwards-compatible: Kubernetes automatically translates them to core/v1, so existing consumers that read events via the old API continue to work unchanged. The multicluster-global-hub agent's CloudEvents emitter continues to function correctly thanks to this translation, but it should be updated to parse the events.k8s.io/v1 API directly to take advantage of the richer field set (action, regarding, related).

Kuttl assertions are updated to remove stale core/v1 fields (reportingInstance, source) and to add events.k8s.io/v1 assertions that verify action, note, regarding, related, and reportingController.

Reference:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/383-new-event-api-ga-graduation/README.md#backward-compatibility


controllers/events.go
```diff
@@ -1,16 +1,41 @@
...
+	"k8s.io/client-go/tools/reference"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// This is the relationship between the CGU Event Reasons and the CGU Event Actions:
+// CguCreated (Reason)
+// - BuildingRemediationPlan (Action) - When a new ClusterGroupUpgrade has triggered the reconciliation.
+// CguStarted (Reason) - When the CGU's enable field is true.
+// - PoliciesRemediationStarted (Action): When the policies remediation is started for the whole ClusterGroupUpgrade.
```
jc-rh: We have started support manifestwork internally. Let's not include "policies" in these fields

greyerof: [Addressed ](https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/10086/changes/609c3e58e5afd29ad737cc527b12f9e7fcf2a9fd#diff-15278cbb719aa3e4a4b04776ec6d9e29d01e623f3bbf8736b46539b7b504536eR51)in commit https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/10086/commits/609c3e58e5afd29ad737cc527b12f9e7fcf2a9fd.

jc-rh (approved):

PR #10171, opened 2026-07-01
Deprecate unused PreCachingConfig override fields

by2waysprojects:
Mark platformImage, operatorsIndexes, and preCacheImage as deprecated in the PreCachingConfig CRD. The controller now derives these values automatically: platformImage from ClusterVersion policies, operatorsIndexes from CatalogSource objects, and preCacheImage from the PRECACHE_IMG environment variable (CSV relatedImages).

Remove the override fallback logic in getPrecacheImagePullSpec() and includePreCachingConfigs() so these fields are no longer read at runtime. Clean up unused function parameters left behind after the simplification.


jc-rh (approved):

tests/kuttl/cgu/pre-caching-with-configs/00-setup-with-pre-caching-config-missing.yaml
```diff
@@ -16,6 +18,14 @@ commands:
   - command: ../../../../deploy/upgrades/pre-caching/patch-policies-status.sh default default default default
     ignoreFailure: false
 
+  # Patch policy1 (ClusterVersion) as Compliant for all spokes
+  - command: >
```
jc-rh, line 22: Should be part of the script above on line 16/18

jc-rh: Never mind. I see it's used by other precaching tests

PR #10175, opened 2026-07-01
Replace text/template with structured object construction for precache CM

by2waysprojects:
Replace the MngClusterActCreatePrecachingSpecCM text template with a buildPrecacheSpecConfigMapAction() function that constructs the ManagedClusterAction as a structured Go object. User-supplied values are assigned as map entries rather than interpolated into raw YAML, making the construction more robust when inputs contain special characters.


jc-rh:
kuttl precache test probably needs an update

jc-rh (approved):

PR #10959, opened 2026-08-07
Detect missing ManagedClusterSetBindings and surface in CGU status

jc-rh:
Add checkPlacementsSatisfied to detect when CGU-owned Placements are blocked by missing ManagedClusterSetBindings (NoManagedClusterSetBindings reason on the PlacementSatisfied condition). When detected, set Reason=WaitingForPlacement on the Progressing condition with a descriptive message guiding users to create the binding; restore Reason=InProgress once resolved.

Also migrate Placement API usage from unstructured.Unstructured to typed clusterv1beta1.Placement for type safety and compile-time validation:
- Rewrite newBatchPlacement, ensureBatchPlacement, getPlacements, updatePlacementWithClusters, cleanupPlacements, checkPlacementsSatisfied and Get/SetPlacementClusterNames to use typed Placement objects
- Widen checkDuplicateChildResources to accept client.Object
- Remove unused PlacementGVK constant
- Register clusterv1beta1 types in scheme and test scheme
- Update all Placement-related unit tests for typed API


controllers/utils/policy_util.go
```diff
@@ -248,116 +247,24 @@ func UpdateManagedPolicyNamespaceList(policyNs map[string][]string, policyNameAr
...
-		for i, name := range clusterNames {
-			values[i] = name
-		}
-	} else {
-		values = []interface{}{}
-	}
-
-	expr["values"] = values
-	return nil
+	placement.Spec.Predicates[0].RequiredClusterSelector.LabelSelector.MatchExpressions[0].Values = clusterNames
```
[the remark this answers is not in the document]
jc-rh: No change needed. All Placements flowing through these functions are created by newBatchPlacement, which always constructs the full predicate/matchExpression structure. The empty-return guards are just nil-safety for the typed API,  not a silent-failure risk.

controllers/clustergroupupgrade_controller.go
```diff
@@ -635,6 +635,29 @@ func (r *ClusterGroupUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.
 					return
 				}
 
+				if clusterGroupUpgrade.RolloutType() == ranv1alpha1.RolloutTypes.Policy {
+					placementMsg, placementErr := r.checkPlacementsSatisfied(ctx, clusterGroupUpgrade)
+					if placementErr != nil {
+						r.Log.Info("[checkPlacementsSatisfied] Error checking placements", "error", placementErr)
```
irinamihai: This should probably be an error log.

controllers/policy.go
```diff
@@ -592,19 +565,44 @@ func (r *ClusterGroupUpgradeReconciler) getPlacements(ctx context.Context, clust
...
+
+	for _, ns := range targetNamespaces {
+		placements, err := r.getPlacements(ctx, clusterGroupUpgrade, nil, ns)
+		if err != nil {
+			return "", err
+		}
+
+		for _, placement := range placements.Items {
+			for _, cond := range placement.Status.Conditions {
+				if cond.Type == clusterv1beta1.PlacementConditionSatisfied && cond.Status == metav1.ConditionFalse && cond.Reason == "NoManagedClusterSetBindings" {
```
irinamihai: Would it be worth having `NoManagedClusterSetBindings` as a constant, with a reference to the upstream source? Even if the Reasons are not exported by the OCM API, if the upstream implementation changes the Reason string, this would fail silently.
The fix would be easier with the constant defined.

controllers/utils/policy_util.go
```diff
@@ -248,116 +247,24 @@ func UpdateManagedPolicyNamespaceList(policyNs map[string][]string, policyNameAr
...
 // SetPlacementClusterNames sets cluster names in Placement spec.predicates[0].requiredClusterSelector.labelSelector.matchExpressions[0].values
-func SetPlacementClusterNames(placement *unstructured.Unstructured, clusterNames []string) error {
-	expr, err := getPlacementMatchExpression(placement)
-	if err != nil {
-		return err
-	}
-	if expr == nil {
-		return fmt.Errorf("matchExpression not found in Placement")
+func SetPlacementClusterNames(placement *clusterv1beta1.Placement, clusterNames []string) {
+	if len(placement.Spec.Predicates) == 0 || len(placement.Spec.Predicates[0].RequiredClusterSelector.LabelSelector.MatchExpressions) == 0 {
```
irinamihai: If this is entered, then in the calling function, an r.Update() would be run unnecessarily.

controllers/utils/policy_util.go
```diff
@@ -248,116 +247,24 @@ func UpdateManagedPolicyNamespaceList(policyNs map[string][]string, policyNameAr
...
-			return nil, fmt.Errorf("value at index %d is not a string", i)
-		}
-		values[i] = str
+	exprs := placement.Spec.Predicates[0].RequiredClusterSelector.LabelSelector.MatchExpressions
+	if len(exprs) == 0 {
+		return []string{}
 	}
-
-	return values, nil
+	return exprs[0].Values
```
irinamihai: Is it safe to return this value directly? If someone makes a change to the returned value, the following SetPlacementClusterName would change the Placement's internals.

irinamihai (commented):
Some comments plus a few notes:
- `checkPlacementsSatisfied` and `updateChildResourceNamesInStatus` are duplicating namaespace gathering for ManagedPoliciesForUpgrade. Might be worth considering to move it to its own function if a 3rd usage is needed in the future.
- Should PlacementBinding move from unstructured to typed `policiesv1.PlacementBinding` as well?

controllers/policy.go
jc-rh: Sure. I did ask claude to use upstream constants. This one is not exported.

jc-rh:
>     * Should PlacementBinding move from unstructured to typed `policiesv1.PlacementBinding` as well?

Yes, we will do it in another PR.

controllers/utils/policy_util.go
jc-rh: Claude code responded to code rabbit with "change not needed" :) , as this is impossible unless someone modifies an TALM owned placement. I updated it anyways.

jc-rh:
coverage collection failure only

